PR #204, opened 2024-11-07
Art harpoon hit vfx

noodlecat519:
Includes all current vfx for when the harpoon hits a surface as prefabs in the PlayerVFX folder. They also exist in the VFXWork scene, which I believe is an art testing gym now.
--VFXHarpoonSparks
--VFXHarpoonSplintersDeck
--VFXHarpoonSplintersTree


ogNINJAGOLDBOX (approved):
Make sure to update from Main

PR #211, opened 2024-11-11
vs-eng-Interaction-System-LV-401

akalscheur-bradley:
- Player has interaction script using Input System
- Works in conjunction with `IPlayerInteractable` interface
- To test, go to the main scene (I didn't make a test gym, oops) and enable `TestInteractable` object under Ferry Boat

(If I need to delete that object from the scene, lmk and I will before I merge)


Assets/GameAssets/Player/PlayerScripts/PlayerFunctionalityCore.cs
```diff
@@ -137,6 +140,11 @@ private void SubscribeToHarpoonInput()
         _harpoonGun.SubscribeInput();
     }
 
+    private void SubscribeToPlayerInteraction()
```
RyanASwanson, line 143: XML comment

Assets/GameAssets/Player/PlayerScripts/PlayerFunctionalityCore.cs
```diff
@@ -177,6 +186,11 @@ private void UnsubscribeToHarpoonInput()
     {
         _harpoonGun.UnsubscribeInput();
     }
+
+    private void UnsubscribeToPlayerInteraction()
```
RyanASwanson, line 190: XML comment

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -0,0 +1,108 @@
...
+
+/// <summary>
+/// Controls the interaction system on the player side
+/// </summary>
+public class PlayerInteraction : MonoBehaviour
+{
+    [Tooltip("How far can the player reach interactable objects")]
+    [SerializeField] private float _maxReach;
+    [Tooltip("Ray that is cast from camera to find interactable objects")]
+    Ray ray;
```
RyanASwanson, line 23: I would prefer if it were specified that this is private, just for reading clarity. Also since it's private it should follow the naming scheme

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    //Input
+    private PlayerInput _playerInput;
+    private InputAction _interactInput;
+
+
+    /// <summary>
+    /// Every frame, Ray is cast.  If the ray hits something and the button is
+    /// </summary>
+    private void Update()
```
RyanASwanson, line 33: I am a coroutine enjoyer and so I would prefer it that way. Not a requirement

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        ray = Camera.main.ViewportPointToRay(new Vector3(0.5f, 0.5f, 0));
```
RyanASwanson, line 35: Perhaps this functionality in here should be in its own functionality. Just would make it easier to read IMO

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -0,0 +1,108 @@
...
+    private void Interact(GameObject gameObject)
+    {
+        if (gameObject.TryGetComponent(out IPlayerInteractable interactable))
+        {
+            interactable.OnInteractedByPlayer();
+        }
+    }
+
+    #region INPUT
+    /*private void OnEnable()
```
RyanASwanson, line 74: Remove the unused functionality

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        SubscribeInput();
+    }
+
+    private void OnDisable()
+    {
+        UnsubscribeInput();
+    }*/
+
+    public void SubscribeInput()
```
RyanASwanson, line 84: XML comment

Assets/GameAssets/World/WorldScripts/Interactables/TestInteractable.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class TestInteractable : MonoBehaviour, IPlayerInteractable
```
RyanASwanson, line 5: Just a reminder to delete the TestInteractable before merging.

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// checks if gameObject can be interacted with
+    ///     If so, calls GiveInteractableFeedback, which should give player feedback that the object is interactable
+    /// </summary>
+    /// <param name="gameObject">GameObject that can potentially be interacted with</param>
+    private void GiveInteractableFeedback(GameObject gameObject)
+    {
+        if (gameObject.TryGetComponent(out IPlayerInteractable interactable))
+        {
+            //REPLACE THIS WHEN WE MAKE UI FOR INTERACTION
+            Debug.Log("Press E to Interact");
```
RyanASwanson, line 57: Perhaps the Interactable could have a function for when it is hovered over so that they can display different things?

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -0,0 +1,108 @@
...
+        _playerInput.currentActionMap.Enable();
+
+        _interactInput = _playerInput.currentActionMap.FindAction("Interact");
+        if (_interactInput != null)
+        {
+            Debug.Log("InteractInput established");
+        }
+        else
+        {
+            Debug.Log("Houston...");
```
RyanASwanson, line 96: I think this is funny, but this should probably specify the issue and could be displayed as a Debug.LogWarning or Error. Imagine a log popping up in console and all it says is Houston... you would be so confused if you didn't write it

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// <summary>
+    /// Every frame, Ray is cast.  If the ray hits something and the button is
+    /// </summary>
+    private void Update()
+    {
+        ray = Camera.main.ViewportPointToRay(new Vector3(0.5f, 0.5f, 0));
+        if (Physics.Raycast(ray, out RaycastHit hit, _maxReach))
+        {
+            GiveInteractableFeedback(hit.collider.gameObject);
+            if (_interactInput.WasPerformedThisFrame())
```
RyanASwanson, line 39: Perhaps this doesn't need to be done in update. I would make the input call a function when pressed and then it checks the status of ray and if that hit something then it calls interact

RyanASwanson (changes requested):
Good start, but could use some cleanup

Assets/GameAssets/Player/PlayerScripts/PlayerFunctionalityCore.cs
ASLH68, line 1: Add self as contributor

Assets/GameAssets/Player/PlayerScripts/PlayerFunctionalityCore.cs
```diff
@@ -23,6 +23,8 @@ public class PlayerFunctionalityCore : MonoBehaviour
     [SerializeField] private PlayerCameraController _playerCamera;
     //Controls harpoon weapon functionality
     [SerializeField] private HarpoonGun _harpoonGun;
+    //Controls player interaction functionality
+    [SerializeField] private PlayerInteraction _playerInteraction;
```
ASLH68, line 27: Instead of serializing, can you run a `GetComponentInChildren` at start

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
ASLH68: Im fine with update being used here, although I do agree it should be in it's own function

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
ASLH68: We need some way for there to be a check if the player is looking at something interactable for UI purposes

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
ASLH68: I think this is fine for now and if we want to add this in later we can

ASLH68 (changes requested):

ASLH68 (approved):

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
RyanASwanson: Yeah I agree the raycast needs to happen every frame, but does the WasPerformedThisFrame need to?

RyanASwanson (approved):
I still feel the if (_interactInput.WasPerformedThisFrame()) doesn't need to be done via update, but regardless I approve

AlecatEnjoyer (approved):
I approve

PR #214, opened 2024-11-11
Vs eng harpoon plays wrong sfx lv 515

ASLH68:
[Jira](https://bradleycapstone.atlassian.net/browse/LV-515?atlOrigin=eyJpIjoiMDc1OGIwM2YzMzE2NGRkZThlZDA1NzgwNmVlYjNmMTUiLCJwIjoiaiJ9)

Minor change so harpoon uses the correct sfx vars


RyanASwanson (approved):
Yeah this changes exactly what you would expect it to change

PR #215, opened 2024-11-11
vs-eng-Fixed-Footsteps-LV-407

ASLH68:
[Jira](https://bradleycapstone.atlassian.net/browse/LV-407)
- Properly released the sound instance
- Removed unnecessary movement toggle code


RyanASwanson (approved):
👍 yup

PR #220, opened 2024-11-12
reordered objects in prefab

JeremiahPeters311:
also cleaned up strange duplicate heart objects that were invisible in scene


hhlaxbro (approved):
Worked aside from the restart but that could be an issue on my end

PR #223, opened 2024-11-13
vs-eng-Interaction-HUD-Prompt-LV-511

akalscheur-bradley:
There is an InteractableTestGym to make sure everything works in Assets/General/Scenes/GeneralTestGyms.
When you look at the box, it brings up the text.  When you look away, the text goes away.  This test gym can also be used for future interactables


Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -21,11 +21,22 @@ public class PlayerInteraction : MonoBehaviour
...
     //Input
     private PlayerInput _playerInput;
     private InputAction _interactInput;
 
+    /// <summary>
+    /// Right away finds the InteractionUI object for future use and disables it before player can see
+    /// </summary>
+    private void Awake()
+    {
+        _interactUI = GameObject.Find("InteractionUI");
```
ASLH68, line 36: I would prefer the UI code to be in its own script and can listen for an event that happens when the player looks at and away from the interactable

ASLH68 (changes requested):

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
ASLH68: Especially in the event that the Interaction UI code eventually will need to be more complicated than a simple enable and disable object, having a system in place that handles an event when the player looks at an interactable would be beneficial in the long run

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -62,20 +73,29 @@ private void SetRaycast()
...
-            //REPLACE THIS WHEN WE MAKE UI FOR INTERACTION
-            Debug.Log("Press E to Interact");
+            if (!_interactUI.activeInHierarchy)
+            {
+                _interactUI.SetActive(true);
+            }
+        }
+        else
+        {
+            if (_interactUI.activeInHierarchy)
```
RyanASwanson, line 90: Why are the else and if separate? Why not just do an else if. 
Or could this be simplified down into not even needing it? Perhaps we could just say "_interactUI.SetActive( (the get component && the !_interactUI.activeInHierarchy)" Just spitballing here

Assets/GameAssets/World/WorldScripts/Interactables/TestInteractable.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System.Collections;
+using System.Collections.Generic;
+using Unity.VisualScripting;
+using UnityEngine;
+
+/// <summary>
+/// class for interactable test
+/// </summary>
+public class TestInteractable : MonoBehaviour, IPlayerInteractable
```
RyanASwanson, line 17: Since this is a test, don't forget to remove it when you are done with it.

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -62,20 +73,29 @@ private void SetRaycast()
 
     /// <summary>
     /// checks if gameObject can be interacted with
-    ///     If so, calls GiveInteractableFeedback, which should give player feedback that the object is interactable
+    ///     If so, makes interactable text appear in HUD.  Otherwise disables text
     /// </summary>
     /// <param name="gameObject">GameObject that can potentially be interacted with</param>
     private void GiveInteractableFeedback(GameObject gameObject)
```
RyanASwanson, line 79: could the name be more specific that gameObject? Because gameObject can also refer to the gameobject this script is attached to

RyanASwanson (changes requested):
Super minor adjustments. Gotta be nitpicky, sorry.

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
akalscheur-bradley: I made the change to reflect else if.  The purpose of the nested if statement for each case is to not unnecessarily call SetActive when the object is already in the state I want it to be in, hence why the 2nd option doesn't work.  See the forwarded message in the PR thread

Assets/GameAssets/World/WorldScripts/Interactables/TestInteractable.cs
akalscheur-bradley: Should it be deleted if it can still be used in the test gym for future interactables? /genq

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
akalscheur-bradley: - The function already uses the name `interactable`.  
- The object is not necessarily interactable, that's why we're testing it.  That means it doesn't necessarily make sense to name it with anything that says `interactable`.
- Anything to the effect of `potentiallyInteractableGameObject` seems unnecessarily wordy
I'm not opposed to renaming it, but I don't have any better ideas

RyanASwanson: Although I agree with your points, I still think we could use a different name than gameObject.

Assets/GameAssets/World/WorldScripts/Interactables/TestInteractable.cs
RyanASwanson: Good question, not entirely certain, but I personally would delete it.

RyanASwanson (approved):

Assets/GameAssets/World/WorldScripts/Interactables/TestInteractable.cs
AlecatEnjoyer: always delete now then bring back later since it will always be saved in github somewhere

AlecatEnjoyer (approved):

RyanASwanson (approved):
I already approved this but I got an @ so take another approval 😜

ASLH68 (approved):

PR #226, opened 2024-11-14
vs-eng-ammo-system-lv-414

astapay:
Ammo Rack is in-engine.
You can test it in PlayerAmmoRackGym. The Ammo Rack has 3 harpoons on it, so you should only be able to shoot 6 times in total.
Jira Task: https://bradleycapstone.atlassian.net/browse/LV-414?atlOrigin=eyJpIjoiMDQ3MjY3ZjFiNWI2NGE3Nzk5NzMzMWE4OWI3NzI5MTYiLCJwIjoiaiJ9
Documentation: https://docs.google.com/document/d/1qyV8T-uRBneLjsCzB8W84C3wD3FJ1SjfEBJIeWANhbo/edit?usp=sharing


Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
```diff
@@ -248,23 +254,36 @@ private void StartReloadProcess()
...
+            {
+                PlayerManager.Instance.InvokeOnHarpoonStartReloadEvent();
+                float reloadTimeRemaining = _reloadTime;
+                while (reloadTimeRemaining > 0)
+                {
+                    reloadTimeRemaining -= Time.deltaTime;
+                    yield return null;
+                }
+
+                _currentReserveAmmo--;
```
RyanASwanson, line 276: Perhaps the reserve ammo could be decremented after firing rather than after reloading. That just makes more sense to me

Assets/GameAssets/World/WorldScripts/Interactables/AmmoRackInteractable.cs
```diff
@@ -0,0 +1,55 @@
...
+    // The nuumber of harpoons that are currently on the rack
+    private int _currentHarpoons;
+
+    /// <summary>
+    /// Called when the game starts
+    /// Sets the number of harpoons to the number of game objects in the ammo rack in-engine
+    /// </summary>
+    private void Awake()
+    {
+        _currentHarpoons = transform.GetChild(0).childCount;
```
RyanASwanson, line 26: This seems odd to me. I feel it would make more sense to input the number of harpoons and have it spawn that many in the barrel rather than the other way arround

Assets/GameAssets/World/WorldScripts/Interactables/AmmoRackInteractable.cs
```diff
@@ -0,0 +1,55 @@
...
+        PlayerManager.Instance.InvokeOnHarpoonRestockEvent(this);
+    }
+
+    /// <summary>
+    /// Removes a harpoon from the ammo rack
+    /// </summary>
+    public void RemoveHarpoon()
+    {
+        _currentHarpoons--;
+        print(transform.GetChild(0).GetChild(0));
```
RyanASwanson, line 43: Don't forget to remove the print statement

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
```diff
@@ -275,14 +294,40 @@ private void HarpoonFullyReloaded()
...
+        if (_harpoonFiringState != EHarpoonFiringState.Ready)
+        {
+            targetAmmo = _maxAmmo;
+        }
+        else
+        {
+            targetAmmo = _maxAmmo - 1;
+        }
+
+        while (!ammoRack.OutOfHarpoons() && _currentReserveAmmo < targetAmmo)
```
RyanASwanson, line 325: More of a preference, but I feel this could be done by just checking the lesser of how many more harpoons can be picked up and how many more harpoons remain and just adding that many. Not a required change, but that makes more sense to me

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
```diff
@@ -275,14 +294,40 @@ private void HarpoonFullyReloaded()
...
+    /// Restocks harpoons to full from the given ammo rack
+    /// </summary>
+    /// <param name="ammoRack"> The ammo rack to take harpoons from </param>
+    private void RestockHarpoons(AmmoRackInteractable ammoRack)
+    {
+        int targetAmmo = 0;
+
+        if (_harpoonFiringState != EHarpoonFiringState.Ready)
+        {
+            targetAmmo = _maxAmmo;
```
RyanASwanson, line 318: If the above comment of reducing ammo when firing is addressed then we shouldn't need the if statement with 2 different ammo counts

RyanASwanson (changes requested):
Some adjustments

Assets/GameAssets/World/WorldScripts/Interactables/AmmoRackInteractable.cs
astapay: I thought so too, but the Jira task states that the number of harpoons that are childed to the ammo rack should determine how many ammo pieces you can get from the rack. I'm just doing what was requested of me.

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
astapay: I had it like this for the sake of displaying the correct information. If the gun has full ammo, then it will appear as 1/2 for example (1 shot in the gun, 2 in reserve). Then, once the gun is fired, it should say 0/2 until it reloads. Once the gun reloads, it will say 1/1.
If we decremented this immediately after firing, it would go from 1/2 to 0/1 to 1/1.

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
astapay: See above comment of reducing ammo when firing.

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
```diff
@@ -248,23 +254,36 @@ private void StartReloadProcess()
...
     }
 
     /// <summary>
     /// The process in which the harpoon is reloaded
     /// </summary>
     private IEnumerator ReloadHarpoon()
     {
-        float reloadTimeRemaining = _reloadTime;
-        while(reloadTimeRemaining > 0)
+        while (true)
```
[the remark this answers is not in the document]
astapay: If we did have an intermediary function, would it not have basically the same functionality? We need to constantly check to see if we have restocked ammo since running out, and to do that, we would need the function to be another coroutine that uses a while(true) statement.

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
astapay: If it's all the same to you, I'd like to keep this as is. I'm sitting here trying to come up with another solution, but I think I like this one the best. Although it is taking the harpoons "one by one," I can't think of another way to solve this problem without using a loop similar to this along the way.

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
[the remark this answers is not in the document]
RyanASwanson: Still disagree, but I'm not dying on this hill. I think avoiding the while(true) would be preferable. To argue with your point of it being the same functionality I don't think so. The while true is checking every frame for if the ammo is empty. With an intermediary function we could have that called only after firing or getting more ammo.

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
RyanASwanson: Still disagree, but I can't force you.

Assets/GameAssets/World/WorldScripts/Interactables/AmmoRackInteractable.cs
RyanASwanson: Fair enough

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
```diff
@@ -44,7 +44,7 @@ private void CheckForInteractable()
         SetRaycast();
         if (Physics.Raycast(_ray, out RaycastHit hit, _maxReach))
         {
-            GiveInteractableFeedback(hit.collider.gameObject);
+            //GiveInteractableFeedback(hit.collider.gameObject);
```
RyanASwanson, line 47: If this isn't being used it should be deleted, otherwise it should be uncommented

RyanASwanson (approved):
I approve. Still think some parts should be done differently. There is a commented out code that should either be removed or uncommented. Resolve that

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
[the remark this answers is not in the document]
ASLH68: I dont see why this needs to be a while(true). Ammo only needs to be checked once before the reload is even called. I say scrap the new functionality here and throw a check for the ammo at the start of the coroutine so it just won't run if theres no ammo.

When ammo is collected, there can be a check to reload the harpoon if it was empty when the ammo was collected

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
ASLH68: In the ammo system, there are two numbers that need to be accurately updated

The ammo currently loaded, and the ammo currently in reserve.
When the player shoots, the reserve number *stays the same* since it has not yet been reloaded. It is decremented here since the reserve changes *after* the reload

Assets/GameAssets/Player/PlayerScripts/PlayerInteraction.cs
ASLH68: ^

ASLH68 (changes requested):

AlecatEnjoyer (approved):
I approve

Assets/GameAssets/Player/PlayerScripts/AmmoUI.cs
```diff
@@ -0,0 +1,91 @@
...
+        PlayerManager.Instance.GetOnHarpoonReloadedEvent().RemoveListener(RestoreMainAmmo);
+        PlayerManager.Instance.GetOnHarpoonRestockCompleteEvent().AddListener(RestoreReserveAmmo);
+    }
+
+    /// <summary>
+    /// Updates the UI when the harpoon is fired
+    /// </summary>
+    private void ReduceMainAmmo()
+    {
+        _ammoText.text = "0/" + _ammoText.text.Substring(2);
```
ASLH68, line 72: I believe in the last PR this used a variable?

Assets/GameAssets/Player/PlayerScripts/AmmoUI.cs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /// <summary>
+    /// Updates the UI when the harpoon is reloaded
+    /// </summary>
+    private void RestoreMainAmmo()
+    {
+        _ammoText.text = "1/" + (int.Parse(_ammoText.text.Substring(2)) - 1);
```
ASLH68, line 80: Same here - didn't this use a variable?

ASLH68 (approved):

PR #227, opened 2024-11-14
vs-eng-Implement-Harpoon-Sfx-Code-LV-466

AlecatEnjoyer:
This branch includes the water splashing sfx from the harpoon hitting the water. Mostly very similar to the boat geometry health script


Assets/GameAssets/World/WorldScripts/Boat/WaterHealth.cs
```diff
@@ -0,0 +1,42 @@
...
+/// </summary>
+public class WaterHealth : BaseHealth
+{
+    /// <summary>
+    /// Overrides the base TakeDamage script to play a specific sfx is the
+    /// Water splashes from the harpoon.
+    /// </summary>
+    /// <param name="damage"> amount of incoming damage </param>
+    /// <param name="damageSource"> incoming damage source </param>
+    public override void TakeDamage(float damage, IBaseDamage damageSource)
```
RyanASwanson, line 22: This seems like a really odd way of doing this. Why does the water require health? We have an interactable interface for when objects are hit by the harpoon. That seems like a better way of addressing that.

RyanASwanson (changes requested):
Not personally a fan of how the water is detecting harpoon contact

Assets/GameAssets/World/WorldScripts/Boat/WaterHealth.cs
astapay: I agree. If we do want to use this, then why bother making it take damage? We could just not use that variable, no?

astapay (approved):
Address Ryan's comment and we good

Assets/GameAssets/World/WorldScripts/Boat/WaterHealth.cs
AlecatEnjoyer: this way was addressed by andrea as a fast way to do it for now that we will change later

PR #235, opened 2024-11-14
Wooden Sparks VFX

AlecatEnjoyer:
This branch belongs to the wooden sparks vfx script which can be applied similar to a sfx script on to a gameobject and it activates.


ASLH68 (approved):

Assets/GameAssets/World/WorldScripts/Boat/WoodenBoatHealth.cs
```diff
@@ -0,0 +1,41 @@
...
+/// Manages the VFX of hitting the boat with the harpoon
+/// </summary>
+public class WoodenBoatHealth : BaseHealth
+{
+    /// <summary>
+    /// Overrides the base TakeDamage script to play the Wooden Spark VFX
+    /// </summary>
+    /// <param name="damage">amount of incoming damage</param>
+    /// <param name="damageSource">incoming damage source</param>
+    public override void TakeDamage(float damage, IBaseDamage damageSource)
```
RyanASwanson, line 21: We have an interface for when a harpoon interacts with something. I feel that makes more sense to use that rather than health. Leaving the same comment on the other task

RyanASwanson (approved):
Approved but should be updated to the harpoon interaction interface later

PR #236, opened 2024-11-14
Metal Sparks VFX

AlecatEnjoyer:
This branch contains the VFX for when the harpoon hits the metal part of the boat. Right now it is only the script only because the boat hasn't been separate to have wood and metal parts yet.


Assets/GameAssets/World/WorldScripts/Boat/MetalBoatHealth.cs
```diff
@@ -0,0 +1,41 @@
...
+/// Manages the VFX of hitting the boat with the harpoon
+/// </summary>
+public class MetalBoatHealth : BaseHealth
+{
+    /// <summary>
+    /// Overrides the base TakeDamage script to play the Wooden Spark VFX
+    /// </summary>
+    /// <param name="damage">amount of incoming damage</param>
+    /// <param name="damageSource">incoming damage source</param>
+    public override void TakeDamage(float damage, IBaseDamage damageSource)
```
RyanASwanson, line 21: We have an interface for when a harpoon interacts with something. I feel that makes more sense to use that rather than health. Leaving the same comment on the other task

RyanASwanson (approved):
Approved but should be updated to the harpoon interaction interface later

astapay (approved):
Hooray!

PR #237, opened 2024-11-15
vs-eng-Functional-submenu-flow-LV-413

JeremiahPeters311:
there is probably going to be a conflict with adam's branch bc i did a similar change to the submenu script there but just use the script from this branch and it should be fine

also minor issue where closing any submenu closes them all at once (I decided I don't have time to get this done right now bc i have a paper to work on as well yay) honestly tho this functionality kinda works better? if we want only closing one at a time, i'll be more than happy to work on it later.


RyanASwanson (approved):
No code changes requested. Some things could be adjusted with the Pause Menu itself. Personally I feel that Controls could be a submenu of settings. In that case you wouldn't even need the menu to choose settings or controls. Options could bring you straight to settings. This is merely a recommendation of what I think would flow best. I would consult with our UI/UX teammates first and get their input,

AlecatEnjoyer (approved):
I approve

SolarFoxG:
There's a bug where menus can be stacked on top of each other as long as the buttons to load them in are unobscured. This specifically affects the pause and main menus; any amount of options menus can be spawned in at once.

menuBug

This can also lead to a fun secondary bug in which, if you have at least three submenus open and spam escape, you can unpause the game with the menu still active.

menuBug2

Lastly, quitting the game through the menus seems to summon a Controls scene.

ASLH68 (approved):

PR #240, opened 2024-11-15
Ambience Event

AlecatEnjoyer:
This branch contains the ambience event which includes a serialize array for applying sfx to for looping through different background sounds to give off authentic ambience.


Assets/General/Audio/FMODScripts/IntervalAmbienceEvents.cs
```diff
@@ -0,0 +1,63 @@
...
+  /// Struct to hold min and max time for random range
+  /// </summary>
+    public struct FModEventReference
+    {
+        public int MinTimeBetweenEvents;
+        public int MaxTimeBetweenEvents;
+    }
+
+    public FModEventReference FmodEvents { get; set; }
+   [SerializeField] private EventReference[] _ambienceArray;
```
RyanASwanson, line 29: This could potentially use a tooltip. **Not a requirement**

Assets/General/Audio/FMODScripts/IntervalAmbienceEvents.cs
```diff
@@ -0,0 +1,63 @@
...
+/// Handler of random ambient sound effects at random times
+/// </summary>
+public class IntervalAmbienceEvents : MonoBehaviour
+{
+  /// <summary>
+  /// Struct to hold min and max time for random range
+  /// </summary>
+    public struct FModEventReference
+    {
+        public int MinTimeBetweenEvents;
```
RyanASwanson, line 24: Can these accessed via { get; set; } ?

Assets/General/Audio/FMODScripts/IntervalAmbienceEvents.cs
```diff
@@ -0,0 +1,63 @@
...
+
+   private Coroutine _ambienceCoroutine;
+   /// <summary>
+   /// Start coroutine 
+   /// </summary>
+   private void Start()
+   {
+       if (_ambienceCoroutine == null)
+       {
+           StartCoroutine(PlayIntervalAudio());
```
RyanASwanson, line 39: You have the _ambienceCoroutine variable but it is never set. This line should be 
_ambienceCoroutine = StartCoroutine(PlayIntervalAudio());

Assets/General/Audio/FMODScripts/IntervalAmbienceEvents.cs
```diff
@@ -0,0 +1,63 @@
...
+            float Timer = 0.0f;
+            for (int i = 0; i < _ambienceArray.Length; i++)
+            {
+                if (Timer > Random.Range(FmodEvents.MinTimeBetweenEvents, FmodEvents.MaxTimeBetweenEvents))
+                {
+                    RuntimeSfxManager.APlayOneShotSfx?
+                        .Invoke(_ambienceArray[i], gameObject.transform.position);
+
+                }
+                Timer += Time.deltaTime;
```
RyanASwanson, line 59: I think you are missing a Yield return null

Assets/General/Audio/FMODScripts/IntervalAmbienceEvents.cs
```diff
@@ -0,0 +1,63 @@
...
+/// <summary>
+/// The Coroutine to loop through the array with different timers per sound call
+/// </summary>
+/// <returns></returns>
+   private IEnumerator PlayIntervalAudio()
+    {
+        while (true)
+        {
+            float Timer = 0.0f;
+            for (int i = 0; i < _ambienceArray.Length; i++)
```
RyanASwanson, line 51: The way this is setup confuses me. To translate this whole thing to pseudocode: In an infinite coroutine loop we are iterating through every value in the ambience array. Then we are checking if that value is more than a random value between the min and max time between events. We are then playing the sound effect at the location of this script. This leaves me with a few questions. Wouldn't it just reset the timer to 0 every frame so it would never actually get to the point of the timer being more than the range? Why are we creating a new random number every frame for each value in the array?

Assets/General/Audio/FMODScripts/IntervalAmbienceEvents.cs
```diff
@@ -0,0 +1,63 @@
...
+// Creation Date:   11/14/2024
+//
+// Description:     Handler of random ambient sound effects at random times
+******************************************************************************/
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using FMODUnity;
+using Random = UnityEngine.Random;
```
RyanASwanson, line 13: I didn't know you could specify the random like this. This is neat. No changes, just good to know :D

RyanASwanson (dismissed):
A good number of suggestions, but the PlayerIntervalAudio coroutine does leave me with some questions

ASLH68 (approved):

Assets/General/Audio/FMODScripts/IntervalAmbienceEvents.cs
```diff
@@ -0,0 +1,63 @@
...
+  /// <summary>
+  /// Struct to hold min and max time for random range
+  /// </summary>
+    public struct FModEventReference
+    {
+        public int MinTimeBetweenEvents;
+        public int MaxTimeBetweenEvents;
+    }
+
+    public FModEventReference FmodEvents { get; set; }
```
ASLH68, line 28: This should be inside the struct

Assets/General/Audio/FMODScripts/IntervalAmbienceEvents.cs
ASLH68, line 29: This should be a list of the Struct you made, not of `EventReference`

ASLH68: The idea is that the struct holds the event reference, and the min/max time between play times for the sfx

astapay (approved):
The others got everything

ASLH68 (approved):

PR #244, opened 2024-11-16
vs-des-HarpoonAnimations

ogNINJAGOLDBOX:
(No Jira Task for it)

Added a bunch of animations and changed the folder layout of the animations.

You can't really test them yet as they need to be programmed. If you want to see them simply attach the Harpoon Animator to the harpoon weapon in PlayerKit and you can scroll through the different animation in the animator

Essentailly here, just make sure the game is still functioning like normal and look at the animations if you would like


JamesBondMeow (approved):
Main scene still runs fine with the new changes implemented

PR #247, opened 2024-11-17
Vs des boatscalingfix

BreadReaver:
Rescaled all the parent objects so now everything should be operating on the same scaling stuff


C-Low01 (approved):
Looks good but fix the box thing 👍

PR #252, opened 2024-11-18
Vs des add map shift beat

Dusk553:
Here is the map shift beat. I was wondering if you would like to look at it in the story beats in the event shifter. It should work, but I was curious if you wanted it in for the milestone


pmetcalf2 (approved):
Boat is shifting when it needs to

PR #254, opened 2024-11-19
ib-eng-designer-friendly-story-beat-system-lv-583

Titonfish:


RyanASwanson (approved):
I approve. I think it's a tad odd to have a boolean for triggering on start. I feel it would just be easier to play the first beat, but I'm assuming that this is for design to start testing on certain beats

PR #259, opened 2024-11-19
Vs des generator story beat

SymphonicSweet:
functionality with generator working, need to disable generator functionality for beat 5 after John's merge


pmetcalf2 (approved):
looks good, beats move on when the generator is interacted with and it cannot progress to beat 5 without the spanner and interacting with the generator

PR #271, opened 2024-11-20
Vs des dialogue implementation

pmetcalf2:
Added Dialogue to Beat 3, Beat 4, and Beat 5

There's delays set up inside of the beats so check  to make sure all dialogue plays

Known bug: when skipping beats with U, the text does not get skipped from previous beats and overrides each other.


JamesBondMeow (approved):
All 3 dialogues work correctly

PR #275, opened 2024-11-21
Vs des boatfixing

BreadReaver:
Bouncing should be reduced to a minimum


Dusk553 (approved):
Everything looks in tip top shape now

PR #289, opened 2024-11-21
Vs eng menu fixing lv 632

JeremiahPeters311:
-fixed main menu size to properly fill the screen

-greyed out settings that don't work

-added numpad enter for title screen

-fixed the infamous enter button bug by removing the default unity input ui "submit" input
nope actually that didn't work because the default input is read only lol
actual solution was just removing the field in event system


hhlaxbro (approved):
Enter bug is gone. I approve

PR #293, opened 2024-11-21
vs-eng-Update-Harpoon-Childing-LV-482

akalscheur-bradley:
A harpoon is childed to the Moving Objects parent object if, when fired, the player is looking at anything under the Moving Objects hierarchy


Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonProjectileMovement.cs
```diff
@@ -16,6 +16,14 @@
 /// </summary>
 public class HarpoonProjectileMovement : MonoBehaviour
 {
+
+    private Transform _movingObjects;
+
+    private void OnEnable()
+    {
+        _movingObjects = FindObjectOfType<BoatMover>().gameObject.transform;
```
ASLH68, line 24: Can this be found on Start or awake since it only needs to be cached once?

ASLH68 (approved):

PR #295, opened 2024-11-21
vs eng footstep audio bug lv 523

ASLH68:
Foot stops properly play and stop when the game is paused and unpaused, and loaded/unloaded


Assets/General/Scripts/Managers/UniversalManagers/MainUniversalManagers/RuntimeSfxManager.cs
```diff
@@ -67,8 +74,12 @@ protected override void SubscribeToGameplayEvents()
     protected override void UnsubscribeToGameplayEvents()
     {
         base.UnsubscribeToGameplayEvents();
+        
         PlayerManager.Instance.GetOnMovementStartEvent().RemoveListener(PlayFootSteps);
         PlayerManager.Instance.GetOnMovementEndEvent().RemoveListener(StopFootsteps);
+        GameStateManager.Instance.GetOnGamePaused().RemoveListener(PauseFootsteps);
+        GameStateManager.Instance.GetOnGameUnpaused().RemoveListener(ResumeFootsteps);
```
RyanASwanson, line 81: Something I noticed here is that this is happening through the game state manager. That confuses me. I see that there are events for game paused/unpaused in the Time Manager. Not sure but I guess at some point that functionality was moved to the GameStateManager. 
This probably isn't something for this PR but we should probably not have that duplicated functionality.

Also there is actually a way to do this without needing 2 separate functions even though the event doesn't take in any variables. I will post an example code in the discord.

RyanASwanson (approved):
Left some suggestions, but I approve

PR #296, opened 2024-11-22
Shooting through walls, again

DavidHenvick:
https://docs.google.com/document/d/17IBmEqQn4UrXPCcg4uYetmu7Pxi-ozqtWAzAMIJJaFo/edit?tab=t.0

code changed in HarpoonProjectileMovement
easy to test by aiming through the porthole at one of the first enemies and moving so the wall is between you and the enemy, and shoot. harpoon is set as inactive after hitting wall.


RyanASwanson (approved):
No issues here

ASLH68 (approved):

PR #299, opened 2024-11-22
vs-bug-fix-player-controller

MilesRogers3D:
Changes:

- Edited values of Cinemachine camera
- Decoupled camera and player so cinemachine's targeting isn't fighting the object's parenting
- Disabled gravity in player Rigidbody and added a manual downward velocity in the movement routine
- Increased gravity
- Added a check for if the player is grounded + not moving to prevent sliding off objects like stairs
- Added a ramp collider for the stairs

### DIFF
--- a/Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
+++ b/Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
@@ -124,8 +124,6 @@ public class HarpoonGun : MonoBehaviour
 
         SubscribeToEvents();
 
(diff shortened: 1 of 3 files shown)


Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
ASLH68, line 1: Add self as contributor

Assets/GameAssets/Player/PlayerScripts/PlayerMovementController.cs
ASLH68, line 1: Add self as contributor

PR #300, opened 2024-11-22
fixed collider not being right on captain quarter

BreadReaver:


C-Low01 (approved):
Could not walk through the captains cabin. Good job :)

PR #302, opened 2024-11-22
Vs eng timed dialogue bug lv 650

nrice25:
- Added a function to prevent the dialogue from ever overlaying
- This function is public so it should be able to get called at any time


Assets/General/Scripts/ScreenText/DialoguePopUps.cs
```diff
@@ -70,6 +75,16 @@ private IEnumerator DisplayText(ScriptableDialogueUI moreDialogue)
         _dataPointer = 0;
         yield return new WaitForSeconds(2f);
         _textContainer.text = ""; 
+        _playingDialogue = null;
+    }
+
+    public void StopDialogue()
```
ASLH68, line 81: XML Comment

ASLH68 (approved):

PR #306, opened 2024-11-23
vs-eng-dialogue-stacking-when-skipping-through-beats-lv-624

Titonfish:


AlecatEnjoyer (approved):

ASLH68 (approved):

Assets/General/Scripts/Story/StoryManager.cs
```diff
@@ -70,6 +82,74 @@ private void Start()
...
     }
 
+    /// <summary>
+    /// Handle any existing and unnecessary or pending coroutines
+    /// </summary>
+    private void FixedUpdate()
+    {
+        PurgeEmptyCoroutines();
+
+        StartPendingCoroutines();
```
RyanASwanson, line 92: To me this seems like quite a bit to have to do in fixed update. Is there no other way?

RyanASwanson (commented):
Personally not a fan that this is all being done in fixed update.

PR #309, opened 2024-11-23
Footstep adjustments

ASLH68:
- Added a short delay before the first footstep sfx plays
- slowed the footstep speed


Assets/General/Scripts/Managers/UniversalManagers/MainUniversalManagers/RuntimeSfxManager.cs
```diff
@@ -192,9 +192,13 @@ private IEnumerator LoopFootSteps()
     {
         float timer = 0.0f;
 
+        yield return new WaitForSeconds(FmodSfxEvents.Instance.FirstFootstepDelay);
```
RyanASwanson, line 195: This WaitForSeconds could be declared before as the time to wait never changes.

RyanASwanson (approved):
Approved, just left a minor optimization comment

PR #311, opened 2024-11-23
vs-des-Gabe'sFixes

ogNINJAGOLDBOX:
https://bradleycapstone.atlassian.net/browse/LV-640
https://bradleycapstone.atlassian.net/browse/LV-646
https://bradleycapstone.atlassian.net/browse/LV-658
https://bradleycapstone.atlassian.net/browse/LV-656

-Fixed the holster animation
-Made the reload time match with the animation, so you cannot shoot while still reloading
-Adjusted harpoon and health prefabs in the flipped version
-Adjusted the collider for the Harpoons Racks so you cannot get ammo through the wall


BreadReaver (approved):
Looks all good, went through and didnt have any issues

PR #313, opened 2024-11-23
vs-eng-Harpoon-Gun-shifting-on-focus-spam-LV-663

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-663
Fixed the issue of the animation moving the harpoon gun slightly over time. Animators shouldn't really move themselves. So to fix this I moved the animator 1 object up. You would think this is simple and takes 2 seconds. It does not. Animators hate it when you move their hierarchy. So I essentially just redid the animations using the old numbers. 
I did notice that the player doesn't holster the weapon when close to a wall. I remember that being a thing, but it isn't working. For the record I didn't break that, its like that in VS
Still think some of the animation triggers should be swapped out for bools but it's not my place to rework. If you do spam the focus unfocus you can have it where the weapon tries to refocus and then unfocuses after you already let go. Likely related to the issue of the triggers not being un set. That's why I recommend a bool. Anyway the task at hand is all good.

### DIFF
--- a/Assets/GameAssets/Player/PlayerAnimations/Harpoon/Focus.anim
+++ b/Assets/GameAssets/Player/PlayerAnimations/Harpoon/Focus.anim
@@ -38,14 +38,14 @@ AnimationClip:
       m_PreInfinity: 2
       m_PostInfinity: 2
       m_RotationOrder: 4
-    path: 
+    path: VisualsHolder
   m_PositionCurves:
   - curve:
       serializedVersion: 2
       m_Curve:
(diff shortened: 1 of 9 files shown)


ogNINJAGOLDBOX (approved):

ASLH68 (approved):

astapay (commented):
Approved

PR #321, opened 2025-01-28
Alpha-Infinite-Ammo-LV-726

Lazydev114:
Jira link - https://bradleycapstone.atlassian.net/browse/LV-726
Description - Infinite ammo functionality where the player never runs out of ammo. just toggle in and out of the mode
Best tested in the main game scene


Assets/General/Scripts/Console/ConsoleController.cs
```diff
@@ -307,6 +315,36 @@ private void ExitInfiniteFocus()
     }
 
 
+    #endregion
+
+    #region infinite ammo
+
+    private void ToggleInfiniteAmmo()
```
RyanASwanson, line 322: Function header

Assets/General/Scripts/Console/ConsoleController.cs
```diff
@@ -307,6 +315,36 @@ private void ExitInfiniteFocus()
...
+    {
+        if (_isInInfiniteAmmoMode)
+        {
+            _isInInfiniteAmmoMode = false;
+            ExitInfiniteAmmoMode();
+        }
+        else if (!_isInInfiniteAmmoMode)
```
RyanASwanson, line 329: This can just be else. Doesn't need to be else if

Assets/General/Scripts/Console/ConsoleController.cs
```diff
@@ -307,6 +315,36 @@ private void ExitInfiniteFocus()
...
+        {
+            _isInInfiniteAmmoMode = true;
+            EnterInfiniteAmmoMode();
+        }
+    }
+
+    private void EnterInfiniteAmmoMode()
```
RyanASwanson, line 336: Function XML

Assets/General/Scripts/Console/ConsoleController.cs
```diff
@@ -307,6 +315,36 @@ private void ExitInfiniteFocus()
...
+    {
+        DoIHaveInifinteAmmo = true;
+        _infiniteAmmoFeedbackText.text = "Exit Infinite Ammo";
+    }
+
+    private void ExitInfiniteAmmoMode()
```
RyanASwanson, line 342: Function XML

Assets/General/Scripts/Console/ConsoleController.cs
```diff
@@ -53,10 +54,16 @@ public class ConsoleController : MonoBehaviour
     private GameObject _spawnedFreeLookCam;//the free look cam that is spawned once the the player turns on free
     //look cam mode
 
+    //infinite ammo mode settings
+    private bool _isInInfiniteAmmoMode = false;
+    [SerializeField] private Button _infiniteAmmoModeButton;
+    public bool DoIHaveInifinteAmmo = false;
```
RyanASwanson, line 60: Seems odd to me that there are 2 booleans for effectively doing the same thing. What is the difference between _isInInfiniteAmmoMode and DoIHaveInfiniteAmmo? I see one is public and one is not, but then why need the private one?

Assets/GameAssets/Player/PlayerScripts/Harpoon/HarpoonGun.cs
```diff
@@ -265,7 +265,8 @@ private void StartReloadProcess()
     /// </summary>
     private IEnumerator ReloadHarpoon()
     {
-        if (_currentReserveAmmo > 0)
+        //nabii added infinite ammo functionality here
+        if (_currentReserveAmmo > 0 || ConsoleController.instance.DoIHaveInifinteAmmo)
```
RyanASwanson, line 269: One problem I noticed is if you are already out of ammo and then activate the debug tool then it doesn't work. Maybe have it so after activating the debug tool it starts the reload?

RyanASwanson (changes requested):
Could use some adjustments to adhere to coding standards. Also there is an oversight when using the debug when out of ammo

RyanASwanson (approved):

PR #322, opened 2025-01-28
Alpha art light effects

noodlecat519:
Created point lights for lanterns and monster eyes. The monster eye lights are in Boss > BossVFX. The lantern lights are in World > WorldArt > WorldEffects.


RyanASwanson (approved):
Approved. It all works as intended.

PR #323, opened 2025-01-28
Lv-728 subtitle background

JeremiahPeters311:
updated script and prefab to make subtitle text have a background. It's a bit of a hacky solution but it works.


Assets/General/Scripts/ScreenText/DialoguePopUps.cs
nrice25, line 1: I see that padding is currently in a string, so you might want to use string interpolation for that; which is using a $ in front of the string so you can put in variables within <> in the string. Either way, looks good

nrice25 (changes requested):
I'm going to wait until the padding and potentially other variables are serialized for design work.

Assets/General/Scripts/ScreenText/DialoguePopUps.cs
```diff
@@ -22,6 +22,30 @@ public class DialoguePopUps : MonoBehaviour
...
+    [SerializeField]
+    private TextMeshProUGUI _textBackgroundContainer;
+
+    [Tooltip("Controls whether or not the text will have a background")]
+    [SerializeField]
+    private bool doTextBackground;
+
+    [Tooltip("background left padding")]
+    [SerializeField]
+    private float leftBackgroundPadding;
```
RyanASwanson, line 35: All these private variables should start with and underscore _

Assets/General/Scripts/ScreenText/DialoguePopUps.cs
```diff
@@ -22,6 +22,30 @@ public class DialoguePopUps : MonoBehaviour
     [SerializeField]
     private TextMeshProUGUI _textContainer;
 
+    [Tooltip("The UI element that renders the text background")]
+    [SerializeField]
+    private TextMeshProUGUI _textBackgroundContainer;
+
+    [Tooltip("Controls whether or not the text will have a background")]
+    [SerializeField]
+    private bool doTextBackground;
```
RyanASwanson, line 31: Booleans should begin with words such as 'does' 'can' 'has'.
Has is probably the best fit here. Also same as the previous comment

Assets/General/Scripts/ScreenText/DialoguePopUps.cs
```diff
@@ -58,23 +82,53 @@ private IEnumerator DisplayText(ScriptableDialogueUI moreDialogue)
...
+            //Debug.Log($"this string should be interpolated{leftBackgroundPadding}");
+
+            //here's where it does text background
+
+            if (doTextBackground)
+            {
+
+                _textBackgroundContainer.text = "<mark=#000000aa padding=�10, 10, 0, 0�>" + dialogueInfo.GetText + "</mark>";
+                //_textBackgroundContainer.text = "$<mark=#000000aa padding=�{10}, 10, 0, 0�>" + dialogueInfo.GetText + "</mark>";
+                //_textBackgroundContainer.text = $"this string should be interpolated{leftBackgroundPadding}";
```
RyanASwanson, line 94: These 2 commented out lines could be deleted. Not going to hold back the push for this, but still worth noting

RyanASwanson (changes requested):
This is really well done. Probably would've taken me a while to figure out this whole padding thing. Just a few notes of things to adhere to the coding standards. After that I will instantly approve (Just @ me)

Assets/General/Scripts/ScreenText/DialoguePopUps.cs
JeremiahPeters311: padding is now serialized! thank you for the string interpolation tip, this is awesome

RyanASwanson (approved):
All good. Hayden Approved too

nrice25 (approved):
Glad you like string interpolation :)

PR #324, opened 2025-01-29
Lv-733-Updating-HUD

hhlaxbro:
Updated the players HUD, including lighting the opacity of the onscreen blood/veins, removing the "reloading" text on reload, and updating the player's heart with new assets.


nrice25 (approved):
All of the UI changes are great improvements

PR #329, opened 2025-01-30
Alpha art low fog shader

noodlecat519:
New fog shader can be found in World>WorldArt>Below Deck Fog>Fog_Textures_Model>ShaderGraph, along with the associated material. This material should be able to be applied to any plane, but if not you can use the circle plane in Fog_Textures_Model>Mesh. If the same material is used on all fog planes, then updating the material on one should update all of them, because that's how Unity materials work. The shader can get data from realtime scene lighting. It shouldn't need any additional code to run when placed in a scene.


RyanASwanson (approved):
Yup that's fog. Certainly fog. Definitely fog. It's fog. 🌫️ fog emoji.

PR #330, opened 2025-01-30
Alpha art light effects

noodlecat519:
Animation for light flickering in World>WorldAnimations. Should be able to be applied to any point light, and won't affect the color of the light. Ryan is planning to write some code so the animation plays at random times.


RyanASwanson (approved):
Yup looks good. May have a merge conflict with your other branch as they both are editing the same scene. Should be an easy one to fix. Also it will require you get this branch up to date with Alpha before merging

PR #337, opened 2025-01-31
Alpha eng light shift horror moment lv 834

astapay:
Light Shift Horror Moment. There is a test gym for this that you can test it in. When you load the scene, there is a little box in the bottom left. When you pass that box, the light shift will trigger.
Jira Task: https://bradleycapstone.atlassian.net/browse/LV-834?atlOrigin=eyJpIjoiNzQ0YzgzNmFmNTgxNDMyOTgyNTY1ZTE0ZTVkNDdlMmMiLCJwIjoiaiJ9
Documentation: https://docs.google.com/document/d/1qyV8T-uRBneLjsCzB8W84C3wD3FJ1SjfEBJIeWANhbo/edit?tab=t.x5ujes8x082n


Assets/GameAssets/World/WorldScripts/HorrorMoments/LightShift.cs
```diff
@@ -0,0 +1,27 @@
...
+// Creation Date :      1/31/25
+//
+// Brief description :  Triggers the Light Shift Horror moment when the player passes through the associated area.
+**********************************************************************************************************************/
+using UnityEngine;
+
+/// <summary>
+/// Activates the Light Shift Horror Moment when passed through
+/// </summary>
+public class LightShift : MonoBehaviour
```
RyanASwanson, line 13: I'm noticing Nabil and you have both made some aspects of your functionality have an OnTriggerEnter check for the player. Going to be odd to have this duplicated. Keep it for now, but keep it in mind. We could make this script more versatile. Rather always playing the vfxmanager event. We could have a serialized event and allow that event allow you to just put in whatever horror moment you want. As the deadline is so close that could be done later if needed as this is perfectly functional.

Assets/GameAssets/World/WorldScripts/HorrorMoments/LightController.cs
```diff
@@ -0,0 +1,144 @@
...
+        float timer = _animator.GetCurrentAnimatorClipInfo(0)[0].clip.length;
+
+        while (timer >= 0)
+        {
+            timer -= Time.deltaTime;
+            yield return null;
+        }
+
+        // Turn off the animator
+        _animator.enabled = false;
```
RyanASwanson, line 124: I am a bit confused on why this is being handled like this. The animator could just set a trigger. I feel this added functionality of enabling and disabling the animator is just unneeded work.

Assets/GameAssets/World/WorldScripts/HorrorMoments/LightController.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        // Confirm we made it back to the original color
+        _light.color = _originalColor;
+    }
+
+    /// <summary>
+    /// Used to play the light flickering animation
+    /// Currently unused because I accidentally made this while doing another task lmao
+    /// </summary>
+    private IEnumerator PlayAnimation()
```
RyanASwanson, line 110: In the context of this script this makes sense. But imagine you were seeing this script for the first time. PlayAnimation is vague. What animation?

Assets/GameAssets/World/WorldScripts/HorrorMoments/LightController.cs
```diff
@@ -0,0 +1,144 @@
...
+    private void LightShift()
+    {
+        // Shift the color
+        StartCoroutine(ShiftColor());
+    }
+
+    /// <summary>
+    /// Keeps track of the light animation and changes the color as needed
+    /// </summary>
+    private IEnumerator ShiftColor()
```
RyanASwanson, line 64: The shifting of color could also be done in animation, but this works too. This is more so just a note rather than a required change

RyanASwanson (approved):
Approved with some notes of how things could be differently. You adhered to the code standards well, and this is well commented. Just noting some alternatives. If you do any adjustments @ me, otherwise this does the functionality that it should so I approve

Assets/GameAssets/World/WorldScripts/HorrorMoments/LightShift.cs
nrice25: Something else to keep in mind, when there are overlaying triggers ~ Unity starts to get weird

Assets/GameAssets/World/WorldScripts/HorrorMoments/LightController.cs
nrice25: Ditto, in the animator you can create a trigger; which can then be called in the script 
```Animator.SetTrigger("DoThing");```
and then set it to a default anim after. 
https://docs.unity3d.com/6000.0/Documentation/ScriptReference/Animator.SetTrigger.html

nrice25 (approved):
Mostly cleanup and agreeing with Ryan on changes that can be done. But, approved

PR #343, opened 2025-02-03
alpha-eng-DuplicatedCameraSway-LV-861

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-861
While trying to fix the issue of the gravity increasing while pausing I found it was fixed and discovered a separate issue. The camera sway from walking would duplicate.


astapay (approved):
Yippee!

PR #346, opened 2025-02-03
Alpha des alpha interactables

pmetcalf2:
made basic prefabs for the other interactables needed for the mazes progression inside of the mazes


BreadReaver (approved):
Looks good everything is pickupable, cant do anything with generator but I assume thats intended.

PR #349, opened 2025-02-04
Lv 870 blockage vine in intro

C-Low01:
Added a placeholder vine to the boat scene that can be destroyed by shooting its weakspot.


BreadReaver (approved):
Works, shoutout

PR #350, opened 2025-02-04
Alpha des monsterimplementation

BreadReaver:
mofster


Caden-Sheahan (approved):
This works for our feedback build!
Monster took damage from the harpoon gun, was able to be shot by it properly using the old weakpoints (placeholder while we're waiting on the final weakpoint system).

PR #351, opened 2025-02-04
Lv 749 intro implementation

SymphonicSweet:
The key can be added to player inventory and once the player interacts with the hatch while the key is in the inventory, the player teleports to where the maze will be. The hatch will not teleport the player without the key in the player inventory.


pmetcalf2 (approved):
hatch is working with the key as intended, good to go ahead and merge

PR #352, opened 2025-02-04
Tuesday build map switch 

pmetcalf2:
made a simple switch that goes from key map to second map when the player picks up the key at the end of the first maze


BreadReaver (approved):
Seems to work all fine on my end

PR #355, opened 2025-02-04
Alpha atmos vfx implimentation

Dusk553:
This is just for the lights. A request so that we can see if the lights don't mess with anything. Don't delete the branch afterwards, I will add in fog from here too.


ogNINJAGOLDBOX (approved):

RyanASwanson (approved):
Lights are in

PR #356, opened 2025-02-04
Lv 785 monster texture implementation

troberts2:
take a look in engine in the "ProceduralTest" scene under bossTestGyms.
Updated model textures and added point lights to the flower eyes.


apollosrw (approved):
looks sick :3

PR #365, opened 2025-02-05
Alpha des fullyshiftingmaze

pmetcalf2:
All map changes up until the ending chase sequence are in this build, check to make sure that all three full mazes and the little post maze 1 section actually change when interacting with the generator.


BreadReaver (approved):
Now this is kinda epic

PR #368, opened 2025-02-05
Updated in scene textures to fix em

BreadReaver:
Updated the mazes in scene to fix some textures so Peyton doesnt have to reload stuff atm


pmetcalf2 (approved):
ceilings no have an actual texture to them now

PR #369, opened 2025-02-05
Alpha eng sound settings lv 730

JeremiahPeters311:
adjusting sound settings now plays a sound clip so you can hear how loud it is

issue: the ambient slider turns on the ambient noises which loops and stacks with itself. I tried to create a version that would stop after a little but fmod is beyond me


astapay (approved):
Hooray!

nrice25 (approved):
There is a bug that should be reported and taken care of later

PR #371, opened 2025-02-05
alpha-eng-Develop-Potty-Break-LV-842

astapay:
Well that was easier than i thought it would be. Test this in PottyBreakGym. All you have to do is take the key off of the toilet.
Jira Task: https://bradleycapstone.atlassian.net/browse/LV-842?atlOrigin=eyJpIjoiNDUzODM4MmQxMDkwNDUzYjhlMzBkZWM4ODhhNjJmNjEiLCJwIjoiaiJ9
Documentation: https://docs.google.com/document/d/1qyV8T-uRBneLjsCzB8W84C3wD3FJ1SjfEBJIeWANhbo/edit?tab=t.8k52n0u8cl47


Assets/GameAssets/World/WorldScripts/HorrorMoments/PottyBreak.cs
nrice25, line 1: This feels like a pretty general script to have that other horror events could use?
If this is only a one and done, then by all means leave it alone;
This is just a nitpick afterall.

nrice25 (approved):
Looks good, small nitpick

PR #372, opened 2025-02-05
Alpha-Scene-Skipping-LV-724

Lazydev114:
JiraLink: https://bradleycapstone.atlassian.net/browse/LV-724
Description: you can now skip between scenes, except for submenu scenes that seem to be a Ui dead zone for some reason. I think it's done on purpose, I will not touch it.


astapay (approved):
Hooray!

nrice25 (changes requested):
Fix the bug causing the game to become unplayable when entering and then trying to leave the main menu from clicking the arrow button

nrice25 (approved):
Approved after removal of submenus from being a part of the tool

RyanASwanson (approved):
I can only read the code as I cannot check in engine. Based on what I have seen in code this is acceptable

PR #373, opened 2025-02-05
Alpha eng creating level interactions lv 784

nrice25:
`Assets\General\Scenes\GeneralTestGyms\LevelInteractions.unity`

- The fan will turn (It currently does not make sound or vfx, but the base is there)
- The radio and telephone can play a sound and have modifiable wait times


RyanASwanson (approved):
I can only look at this through code, not editor but I approve.

PR #377, opened 2025-02-06
Alpha des game ending area lv 846

JamesBondMeow:
Following the jira task, duplicated the game scene and moved the player spawn point onto a patch of grass


C-Low01 (approved):
Get the build thing fixed with the new automated system then you're good.

PR #382, opened 2025-02-06
Alpha des monsterimp

BreadReaver:
Added monsters into maze 2 and 3


pmetcalf2 (approved):
monsters are in the maze, go ahead and merge

PR #396, opened 2025-02-10
Story Beat fixing

SymphonicSweet:
Hatch and player waking up works with the story beat manager, however the vine cannot be destroyed, therefore the next story beat cannot be played.


C-Low01 (approved):
Update your branch with main first, then you are good to merge

C-Low01:
Remember to merge alpha into your branch before merging

PR #399, opened 2025-02-10
tommys overwritten changes from alpha implementation

troberts2:
hopefully everything works again. I really only changed the procedural vine during the implementation sprint, so I just made sure it worked again. The vine can retract when shot in the head and also I changed how the harpoon interacts with triggers so that it doesn't hit invisible triggers we don't want it to.


Assets/GameAssets/Boss/BossScripts/Attacks/ProceduralAttack/Editor/ProceduralVineEditor.cs
```diff
@@ -0,0 +1,25 @@
...
+*****************************************************************************/
+using UnityEditor;
+using UnityEngine;
+/// <summary>
+/// Custom editor class for the vine so I can mess with it without having to use the player
+/// </summary>
+[CustomEditor(typeof(ProceduralVine))]
+public class ProceduralVineEditor : Editor
+{
+    public override void OnInspectorGUI()
```
RyanASwanson, line 16: Please give this an XML comment. That's all

RyanASwanson (approved):
Approved, just asked for 1 XML

PR #400, opened 2025-02-10
Fix-Player-Reload-Being-Tied-to-Other-Objects-LV-883

nrice25:
Decoupled various dependencies from the player

You can try it in the `FixingMovement` scene within the testing scenes.
Various code cleanup happened allowing the player to shoot in scenes that are not the main two


RyanASwanson (approved):
Approved. There are still errors in the console but those have to do with the ConsoleController, and I believe is unrelated to this task.

PR #404, opened 2025-02-11
Alpha eng scene transitions

JeremiahPeters311:
should just be a redo of the previous scene transition commit, most of the stuff here is literally copy pasted from the old pull request on github. As with then, there are no actual animations for transitions to do, but you can test it out with the delay variable which does function properly.


hhlaxbro (approved):
Looks to be good from what I can tell.

nrice25 (approved):
Looks good to me; I think you'll need to wait for Ryan to approve

PR #406, opened 2025-02-11
Dialogue implementation into Story beat system (LV 744)

C-Low01:
That boy is in there!


C-Low01:
All changes can be viewed in Gamescene

PR #407, opened 2025-02-11
alpha-eng-potty-break-reimplementation

astapay:
I did the thing again. It's just Potty Break again, same thing we've seen before.
Documentation: https://docs.google.com/document/d/1qyV8T-uRBneLjsCzB8W84C3wD3FJ1SjfEBJIeWANhbo/edit?tab=t.8k52n0u8cl47


RyanASwanson (approved):
I still feel like this code could be made more general and multiuse but this achieves the goal it is set out to achieve so I will approve

nrice25 (approved):
Code looks good

PR #408, opened 2025-02-11
Alpha-des-HorrorMomentsReshoot

ogNINJAGOLDBOX:
Implemented the Light Change and Slytherin horror moments. Also sort of implemented Thump In The Night but the sound is just the monster screech as the sounds are not in. Also tried starting on the chase sequence, but currently no way to get to the final maze part. Also adjusted the boat sway to be less extreme.

https://bradleycapstone.atlassian.net/browse/LV-776
https://bradleycapstone.atlassian.net/browse/LV-837
https://bradleycapstone.atlassian.net/browse/LV-840

To see the changes, all you need to do is play from the beginning of the maze section.
- First Maze: you will see the lights change to Green at a certain point, then you will hear a screech from the monster later on
- Second Maze: At a certain point you will see a monster down around a corner, when you get close the light above will flicker more than usual and then as you get closer it will slither around the corner and disappear.
- Third Maze: None Yet

Also tried to implement a chase sequence, but can't test it as it is not in the scene and cannot be gotten to yet

Also changes the boat sway to not be so intense

Also added Post Processing


RyanASwanson (approved):
This should be good. Oliver has looked over it and verified no changes should be overwritten to the mazes so I am confident to approve

PR #419, opened 2025-02-12
Alpha des ending sequence

pmetcalf2:
Ending chase sequence added after interacting with the generator, at the end it the hatch that takes you to the ending scene.
ending scene is should be the boat leaving without the player, and the scene transitions to the win screen after a short time.


BreadReaver (approved):
Bit of a corny ending but hey it works

PR #423, opened 2025-02-12
alpha-des-HorrorMomentsTheFinale

ogNINJAGOLDBOX:
Added some sounds, changes footstep repeat length so not so much, Tried adding Potty Break (bugged currently)

https://bradleycapstone.atlassian.net/browse/LV-777
https://bradleycapstone.atlassian.net/browse/LV-838

To see them, play through the maze and you should see:

1. Light Change with sound
2. Thump with slither
3. Tentacle Slither Away down the hall with flickering light sounds and slither sound

Also footsteps aren't as frequent


RyanASwanson (approved):
Approved

PR #432, opened 2025-02-13
Alpha-Scene-Transition-Buttons-LV-940

Lazydev114:
Jira Link: https://bradleycapstone.atlassian.net/browse/LV-940
Description: Added the scene transition button back into the dev-console
Best tested: check if the scene transition button is in the dev console and if the buttons work in the game scene. You may need to exit the dev console for the scene transition to work


RyanASwanson (approved):
Approved. But as noted in discord, it does feel odd that the player does have to close the debug menu before the scene transition actually works

PR #440, opened 2025-02-13
alpha-eng-flower-head-clipping-away-from-vine-model-LV-914

troberts2:
literally just lowered the flower head gameObject closer to the vine to reduce clipping.


RyanASwanson (approved):
Approved. Nice!

PR #442, opened 2025-02-14
Alpha des qa mazefixes

BreadReaver:
Fixed 2 jira bugs
Put ceiling back on transition area
Cleared up some more texture clipping


RyanASwanson (approved):
Looks like the issues were resolved

PR #444, opened 2025-02-14
Mouse does not appear in pause menu-931

JeremiahPeters311:
checked "run in background" in build settings. I could not replicate the bug as described, even when testing extensively in build, but my research indicates this might be a potential fix.


RyanASwanson (approved):
Guess it works now. Huh neat

PR #445, opened 2025-02-14
Alpha des simplifying mazes

BreadReaver:
Simplified all 3 maps based on some concepting work I did


RyanASwanson (approved):
Looks good to me. Maze 3 could still be a tad easier, but this is definitely an improvement

PR #446, opened 2025-02-14
Alpha-eng-Footstep-audio-loops-after-screen-transition-LV-948

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?customFilter=ari%3Acloud%3Ajira-software%3A792ff6f6-f5ba-4734-a1cd-457d34518a06%3Acustom-filter%2Factivation%2F489b3b11-a1a5-4c9e-afed-c8d621aa5e68%2F32%2F0ffdcf80-e13c-47e4-8c44-6b732fc8693c&selectedIssue=LV-948

I think our automated branch naming is wrong. Gotta talk to Andrea about that
This task was like removing 3 characters. Despite that it took a good bit longer than expected. Basically the issue is that the player could technically start moving multiple times without ever stopping the movement. Happened during scene transitions because the framerate was low enough to press and release the input in the same frame easily. Can be tested by spamming the input buttons during a transition. Test it against alpha to be certain.


nrice25 (approved):
Works in engine, simple code change; approved

PR #447, opened 2025-02-14
World Space UI doesn't disappear after reloading-966

JeremiahPeters311:
ammo racks can now call an event when they run out of ammo, used this to disable their own pop up.

I also fixed the key enabling the hatch pop up cuz that broke at some point


Assets/GameAssets/World/WorldScripts/Interactables/AmmoRackInteractable.cs
```diff
@@ -17,6 +19,8 @@ public class AmmoRackInteractable : MonoBehaviour, IPlayerInteractable
     // The nuumber of harpoons that are currently on the rack
     private int _currentHarpoons;
 
+    [SerializeField] private UnityEvent _onAmmoDepletedEvent;
```
RyanASwanson, line 22: Only a minor suggestion. I would recommend having some way to get the event from other scripts. A getter function perhaps? Or make it public and change the naming? Either works. This isn't a required change, just a suggestion

RyanASwanson (approved):
Approved, just a minor suggestion

PR #448, opened 2025-02-14
Alpha-Missing-Prefab-In-The-Maze-Scene-Rework-LV-951

Lazydev114:
Jira Link: https://bradleycapstone.atlassian.net/browse/LV-951
Description: The dev console is now part of the player package. It can be opened wherever the player is. As Ryan said, killing two birds with one stone
This is the other task that it fixes: https://bradleycapstone.atlassian.net/browse/LV-884


RyanASwanson (approved):
Made a tiny commit to this branch, just to be certain that the debug consoles don't duplicate. Aside from that I approve!

PR #453, opened 2025-02-14
Fade in Shorter

pmetcalf2:
The fade in when the game starts is much shorter that's it


JamesBondMeow (approved):
works, a lot faster fade in

PR #454, opened 2025-02-14
Alp eng Slider should play sfx on release LV 973

Titonfish:


RyanASwanson (approved):
Yup. It works. Approved

PR #457, opened 2025-02-15
Switched the sfx to a different one that would not loop (Hopefully)

Dusk553:
Jura task:https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=63c32962c52a63dcda83f747&selectedIssue=LV-953 

The SFX should likely not loop after opening the menu and then closing it


RyanASwanson (approved):
Yup its working now

PR #458, opened 2025-02-15
alpha-eng-Monster-Hitbox-Inconsistency-LV-961

troberts2:
test in procedural test scene.
During my testing it seemed that 1/5 of the times I shot the head it would not detect it. I tried increasing the collision detection mode and also increasing the radius of the heads collider. This fixed all of nothing so I tried googling it and at the cost of some performance you can change the fixed timestep in project settings to a lower number and that seemed to do the trick


RyanASwanson (approved):
Tentative approval.

PR #460, opened 2025-02-15
alpha-eng-Monster-doesn't-deal-damage-after-first-hit-LV-952

troberts2:
test in procedural test. I added my own OnCollisionEnter to PlayerHealth rather than trying to figure out how to use our health system. I'm not sure if objects tagged Enemy were even damaging the player before this.


RyanASwanson (changes requested):
Ok I performed some additional checks and the enemy collision is acting weird

PR #461, opened 2025-02-15
Player spawn point moved

C-Low01:
Moved player spawn point to the front of the boat instead of the back


SymphonicSweet (approved):
Everything works as intended, player spawns in the correct location and does not break the game.

PR #462, opened 2025-02-15
Disappear Interactable UI-967

JeremiahPeters311:
added variable to control distance where interactable pop up disappears from view


Assets/GameAssets/UI/UIScripts/WorldSpace Popups.cs
```diff
@@ -60,7 +64,12 @@ private void LateUpdate()
         if (_playerReference != null)
         {
             //check proximity to player
-            if (Vector3.Distance(_playerReference.transform.position, transform.position) >= _playerDetectionProximity)
+            if (Vector3.Distance(_playerReference.transform.position, transform.position) >= _visibilityProximity)
```
RyanASwanson, line 67: Only thing I can say is it is also an option to store this distance as a float first. That way we don't have to calculate it twice. That doesn't mean more memory is being used though. Honestly it's acceptable either way.

RyanASwanson (approved):
Approved. Nothing to say here. Looks like things are working as intended. Just left a teeny tiny suggestion

hhlaxbro (approved):
Works great!

Assets/GameAssets/UI/UIScripts/WorldSpace Popups.cs
JeremiahPeters311: 👍

PR #463, opened 2025-02-15
Alpha des populate mazes with assets lv 974

Dusk553:
Added in many props to all three of the easier mazes to make them look a little less empty. I ask that you check to make sure nothing breaks from them being added to the prefabs.

Jura Task:https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=63c32962c52a63dcda83f747&selectedIssue=LV-974


RyanASwanson (approved):
Yup those are boxes and barrels. Maybe the barrel doesn't need to be a capsule collider, but that's minor. Approved.

PR #465, opened 2025-02-16
alpha-art-generatorfixes-lv-912 updated generator textures and made generatorv3 prefab 

angelkait:
hiiiii i have made the generator cooler and created a new prefab for it in the same place the old one was! i wasn't sure if anyone had already used the old one so i just left it there instead of deleting it so i wasn't stepping on anyone's toes


RyanASwanson (approved):
One of the textures seems a tad stretched but that's not something to stop the merge over

PR #469, opened 2025-02-16
alpha-des-Increase-Tendril-Size-LV-965

ogNINJAGOLDBOX:
Increased size of monster tendrils

https://bradleycapstone.atlassian.net/browse/LV-965

Just need to go see the monsters, simple prefab change should have changed them all

Might need to adjust the pathing and detection a bit because it clips a lot, but that is another task that is beyond my experience with the monster


RyanASwanson (approved):
Approved. Github is not broken and I am happy

PR #473, opened 2025-02-17
Updated monster positions

JamesBondMeow:
Updated monster positions in engine so monster tendrils fully retract before despawning; Origin lowered so it can fully retract before despawning.

https://bradleycapstone.atlassian.net/browse/LV-947


BreadReaver (approved):
Will talk about getting death clipping fixed up a bit tomorrow. Patrol placement seems good though

RyanASwanson (approved):
I agree with Oliver about the death, but you've planned to discussed that you plan to fix it.

PR #474, opened 2025-02-17
Updated monster positions

JamesBondMeow:
Updated monster positions in engine so monster tendrils fully retract before despawning.

https://bradleycapstone.atlassian.net/browse/LV-947


BreadReaver (approved):
Looks good, worried it may conflict with your other changes though

RyanASwanson (approved):

PR #481, opened 2025-02-19
transition to chase sequence level lv 1001

pmetcalf2:
Added the chase sequence to the story manager in maze scene, extra beats are there for when we add in the monster to the scene


BreadReaver (approved):
Looks good, will probably need to move the monsters around a little bit in the future but playspace is functional fo rnow

PR #482, opened 2025-02-19
beta-eng-Can-use-healing-packs-at-full-health-LV-987

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-987

Updated the health packs to change intractability based on if the player is full health or not. Best tested in the above deck scene. You shouldn't be able to use the health packs. Then open the debug menu (its bound to c) and inflict damage to yourself (you have to set the amount of damage first) Health packs heal 20 health currently, so if you inflict 30 damage you can use it twice.
Engine Review Estimation: <3 minutes


astapay (approved):
Yippee!

PR #483, opened 2025-02-19
beta-eng-Chase-Monster-Attack-Functionality-LV-1003

troberts2:
[Jira](https://bradleycapstone.atlassian.net/browse/LV-1003?atlOrigin=eyJpIjoiMGQ0NGRlMDVlZWFjNDc2ZDk3ZjU1YjAyNTU5OTlhN2IiLCJwIjoiaiJ9)
Test it out in scene "ProceduralTest".

I added each group of vines that chase you as children of the "Ending Chase Sequence" gameObject. NOTE: these are not on updated prefab yet. I did not want to overwrite prefab if work has been done.

You can adjust settings and speed on each gameObject named "GroupOfVinesN". You can also adjust if it shakes the camera when starting, if it instantly kills player, if it's triggered by the player walking through it, and the damage it does.


Assets/GameAssets/Boss/BossScripts/Attacks/ProceduralAttack/ChaseVineGroup.cs
```diff
@@ -0,0 +1,74 @@
...
+/// <summary>
+/// This class controls the group of chasing vines
+/// </summary>
+public class ChaseVineGroup : MonoBehaviour
+{
+    [Header("Chase settings")]
+    [SerializeField] private float _chaseSpeed = 3f;
+    [SerializeField] private Transform _chaseCollider;
+    [SerializeField] private bool _isTriggeredByPlayerWalkingBy = true;
+    [SerializeField] private bool _killPlayerInstantly = true;
```
RyanASwanson, line 19: booleans should begin with words such as: is, can, does, etc.

Assets/GameAssets/Boss/BossScripts/Attacks/ProceduralAttack/ChaseVineGroup.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Makes collider follow the vines when appropriate
+    /// </summary>
+    private void Update()
+    {
+        if(_colliderFollow != null)
+            _chaseCollider.position = _colliderFollow.position;
```
RyanASwanson, line 47: Coding standards state that we shouldn't have if statements like this without the brackets. Should be
if(_colliderFollow != null)
{
            _chaseCollider.position = _colliderFollow.position;
}

astapay (approved):
Yippee!

PR #484, opened 2025-02-20
Beta des texture cptn qrtrs and stairs lv 1031

BreadReaver:
Added a new texture to the captains quarters on the top deck to make it fit in more


pmetcalf2 (approved):
looks way better than in main right now, go ahead and merge

PR #489, opened 2025-02-20
Beta des seamless maze transitions lv 1027

pmetcalf2:
generator moved to the center of all generator rooms and lined them up better when the mazes switch


BreadReaver (approved):
Generator placement looks good, nice work

PR #490, opened 2025-02-20
Key Scaled up and repositioned

MaskedMan152:
Key is bigger and shouldn't be clipping into the wall


SymphonicSweet (approved):
looks good to me, works as intended, it's still a little dark so the key is hard to see, but it is bigger and easier to pick up

PR #491, opened 2025-02-20
beta-eng-Player-Walking-Feels-Bad-LV-976

astapay:
Hooray! The gun now moves left to right instead of the player! Test this in any scene
Jira Task: https://bradleycapstone.atlassian.net/browse/LV-976?atlOrigin=eyJpIjoiMTJkZjdiNTM0NGQyNDRkNThmMmE0NGVjNWFkZjY0YjIiLCJwIjoiaiJ9


Assets/GameAssets/Player/PlayerScripts/PlayerCameraController.cs
```diff
@@ -45,12 +45,12 @@ public class PlayerCameraController : MonoBehaviour
...
+    [SerializeField] private GameObject _harpoonGun;
     [SerializeField, Range(0f, 10f)] private float _movementSwaySpeed = 5f;
     [SerializeField, Range(0f, 10f)] private float _movementSwayIntensity = 5f;
-    private const float _BASE_MOVEMENT_SWAY_SPEED = 0.0025f;
-    private const float _BASE_MOVEMENT_SWAY_INTENSITY = 0.25f;
-    private const float _MOVEMENT_SWAY_SPEED_LIMITER = 5f;
-    private const float _MOVEMENT_SWAY_INTENSITY_LIMITER = 5f;
+    private const string _TARGET_ANIMATION = "harpoonIdle";
+    private const float _BASE_MOVEMENT_SWAY_SPEED = 0.00005f;
+    private const float _BASE_MOVEMENT_SWAY_INTENSITY = 0.004f;
```
RyanASwanson, line 53: Do these things need to be private const? Could they be designer accessible? Probably would need a range to prevent them setting the values too high or low. Not required, but would be nice

Assets/GameAssets/Player/PlayerScripts/PlayerCameraController.cs
```diff
@@ -181,28 +183,21 @@ private IEnumerator WalkingSway(InputAction playerMovement)
...
                 }
-                _transposer.m_FollowOffset = new Vector3(newX, _transposer.m_FollowOffset.y, newZ);
+                _harpoonGun.transform.localPosition = new Vector3(newX, 0, 0);
 
                 // If we reach the limit on our sway, switch directions
-                float limit = _BASE_MOVEMENT_SWAY_INTENSITY * _movementSwayIntensity /
-                    _MOVEMENT_SWAY_INTENSITY_LIMITER;
-                Vector3 origin = new Vector3(0, _transposer.m_FollowOffset.y, 0);
-                float distance = Vector3.Distance(_transposer.m_FollowOffset, origin);
+                float limit = _BASE_MOVEMENT_SWAY_INTENSITY * _movementSwayIntensity;
```
RyanASwanson, line 199: Can we clarify what the limit is in respect to?

Assets/GameAssets/Player/PlayerScripts/PlayerCameraController.cs
```diff
@@ -181,28 +183,21 @@ private IEnumerator WalkingSway(InputAction playerMovement)
...
+                float distance = Vector3.Distance(_harpoonGun.transform.localPosition, Vector3.zero);
```
RyanASwanson, line 200: Can we clarify what the distance is in respect to?

Assets/GameAssets/Player/PlayerScripts/PlayerCameraController.cs
```diff
@@ -164,14 +164,16 @@ private void StartWalkingSway(InputAction playerMovement)
     private IEnumerator WalkingSway(InputAction playerMovement)
     {
         Coroutine stopSwayCoroutine = null;
+        Animator _harpoonAnimator = _harpoonGun.GetComponent<Animator>();
```
RyanASwanson, line 167: Personally I would set this in start. That way if we ever stop this coroutine and start it again we wouldn't need to get the variable again.

Assets/GameAssets/Player/PlayerScripts/PlayerCameraController.cs
```diff
@@ -45,12 +45,12 @@ public class PlayerCameraController : MonoBehaviour
 
     // Variables for movement sway
     [Space]
+    [SerializeField] private GameObject _harpoonGun;
```
RyanASwanson, line 48: Does this even need to be serialized? If it really does, why is it set to a gameobject if we are just using it to get to an animator?

RyanASwanson (approved):
Left some notes on some stuff that could be adjusted, regardless the core functionality looks good.
Some notes: The movement of the weapon feels rather static and mathematical.
And you are going to hate me for this one. Honestly it feels less natural to me without the camera sway, but hey I ain't the one making those choices.

Assets/GameAssets/Player/PlayerScripts/PlayerCameraController.cs
astapay: We're also using it to move the harpoon gun itself, not just to check the animator. Since we need both the Transform and the Animator from the same object, might as well just get the entire thing.

Assets/GameAssets/Player/PlayerScripts/PlayerCameraController.cs
astapay: These are basically constants that limit the movement sway value. They are in place basically to put the sway speed and intensity on a 1-10 scale. I could remove them, but that would require design to make the other values incredibly close to 0.

PR #492, opened 2025-02-20
Beta des implement other chase sequence audio lv 1005

pmetcalf2:
added the generator start sounds to the beginning of the chase sequence


BreadReaver (approved):
Sound is in but gets kind of covered up by the squelchy level transition audio

PR #494, opened 2025-02-21
Beta des implement generator, spanner, hatch lv 1017

SymphonicSweet:
added the generators, hatch, and spanner. Interactability should still be there.


C-Low01 (approved):
Why isnt the generator room light on?

PR #501, opened 2025-02-23
Beta eng voice acting dialogue functionality lv 1157

Titonfish:


RyanASwanson (approved):
Yeah this looks good to me. I've tried swapping in multiple audio tracks for dialogue and I see no issue with that. Looks like its working properly. Only note I have is the documentation could be more filled out in some parts

nrice25 (approved):
Looks good outside of game. In game testing will prove itself later

PR #502, opened 2025-02-23
LV-1188-Streamlining-Maze-3-Runback

BreadReaver:
Made Peyton's crate wall a prefab
Used the crate walls to create a single path back through maze three once you pick up the gascan
Added it to the story manager


pmetcalf2 (approved):
Its much better going back through the maze, go ahead and merge it in

PR #506, opened 2025-02-24
beta-eng-Dead-Vines-Can-Kill-LV-1007

troberts2:
disabled colliders in children so that the player cannot get hit after shooting a vine.


RyanASwanson (approved):
Yup those vines don't kill

PR #519, opened 2025-02-26
beta-eng-ALT-save-functionality

RyanASwanson:
Alternative save functionality based on Nick Rice's work on the save system. Adjusted to be more designer friendly.


astapay (approved):
Yippee!

PR #523, opened 2025-02-27
Beta-Trailer-Mode-Dev-Console-LV-1175

Lazydev114:
JiraLink: https://bradleycapstone.atlassian.net/browse/LV-1175
Description: A mode without ui.
It is best tested in a scene with the player in it


Assets/General/Scripts/Console/ConsoleController.cs
```diff
@@ -475,6 +485,31 @@ private void ToggleDevUiMode()
...
+    /// trailer mode means no ui
+    /// </summary>
+    private void ToggleTrailerMode() 
+    {
+        if (_isInTrailerMode)
+        {
+            //turn player hud off
+            _playerHud.SetActive(true);
+            _trailerModeButtonText.text = "Enter Trailer Mode";
+            _isInTrailerMode = false;
```
RyanASwanson, line 501: We could actually shorten this if else a bit. We could put things outside the if statement and say things such as
_playerHud.SetActive(_isInTrailerMode);
_isInTrailerMode = !_isInTrailerMode

Assets/General/Scripts/Console/ConsoleController.cs
```diff
@@ -73,6 +73,13 @@ public class ConsoleController : MonoBehaviour
     [SerializeField] private TMP_Text _devUiModeText;
     [SerializeField] private GameObject _devUi;
 
+
+    //trailer mode
```
RyanASwanson, line 77: Not required but something I find nice to do for design. We can use [Header("Trailer Mode")]   and that shows up in the inspector. Makes things easier to read. Just a nice little trick

RyanASwanson (changes requested):
Code seems good. Could use some minor adjustments, but regardless I like it. This doesn't currently work in the maze scene so I'm requesting changes.

RyanASwanson (approved):
Yup. Bug was fixed, changes were implemented. Doesn't appear to be breaking anything. Looks good to me.

PR #524, opened 2025-02-27
beta-eng-Monster-Appearing-Proc-Animation-LV-1138

troberts2:
You can test in proceduralTest scene. Go under the "Stem(attack)" gameObject and click on child "followPath". In the inspector you can trigger the appear animation and the retract animation.

It is very buggy currently and after appearing the vine gets shorter. Repeatedly retracting and appearing the vine will make it small and buggy. Only use once currently.


Assets/GameAssets/Boss/BossScripts/Attacks/ProceduralAttack/ProceduralVine.cs
```diff
@@ -51,6 +51,15 @@ public class ProceduralVine : MonoBehaviour
 
     private bool _isRetracting = false;
 
+    [Header("Appear Stuff")]
+    [SerializeField] private PathCreator _appearPath; // The target to move toward
+    [SerializeField] private float _appearSpeed = 5f; // Speed of movement
+    [SerializeField] private float _appearDistance = 0;
+
+    private bool _isAppearing = false;
+    private bool _isShftingBack = false;
```
RyanASwanson, line 60: We could switch these 2 bools out for an enum to track the state of the vine. Not a required change, but generally when I see multiple bools for the state something is in my brain goes to enums as a solution. 
To note: Enum declarations should always start with E. So for example. EVineState

Assets/GameAssets/Boss/BossScripts/Attacks/ProceduralAttack/ProceduralVine.cs
```diff
@@ -74,15 +83,41 @@ private void Update()
...
+        }
+    }
+
+    private IEnumerator JumpBackToPath(float timeToGetToPath)
+    {
+        _distance = 0;
+        _followTransform.DOJump(_pathCreator.path.GetPointAtDistance(_distance), .2f, 1, timeToGetToPath, false).SetEase(Ease.OutQuad);
+        Vector3 direction = (_pathCreator.path.GetPointAtDistance(_distance) - _followTransform.position).normalized;
+        _followTransform.forward = direction;
+        yield return new WaitForSeconds(timeToGetToPath);
```
RyanASwanson, line 119: A minor optimization we learned a while back is that for coroutines that only use WaitForSeconds we can replace those with a regular function that uses PrimeTween.Tween.Delay as a replacement. The primetween method creates less garbage. This is not a required change as it is a miniscule optimization, but if you would like some help let me know

Assets/GameAssets/Boss/BossScripts/Attacks/ProceduralAttack/ProceduralVine.cs
```diff
@@ -218,4 +254,29 @@ private void Retracting()
...
+    #region Appear
+    /// <summary>
+    /// starts the vine appear
+    /// </summary>
+    public void StartAppear()
+    {
+        _appearDistance = 0;
+        _isAppearing = true;
+        _chainIKRig.weight = 0;
+        _dampedTransformRig.weight = 1;
```
RyanASwanson, line 268: Could we include some more comments in this. To an outsider some of these changed variables might be hard to understand what they are doing. This generally applies to this script as a whole. To you the procedural animation stuff makes sense, but to everyone else its more difficult, so comments could help.

Assets/GameAssets/Boss/BossScripts/Attacks/ProceduralAttack/ProceduralVine.cs
```diff
@@ -51,6 +51,15 @@ public class ProceduralVine : MonoBehaviour
 
     private bool _isRetracting = false;
 
+    [Header("Appear Stuff")]
+    [SerializeField] private PathCreator _appearPath; // The target to move toward
+    [SerializeField] private float _appearSpeed = 5f; // Speed of movement
```
RyanASwanson, line 56: I've noticed the vine gets really weird when put at higher speeds. The default seems to be 1.5 in editor, I tried changing it to 5 and it got all clustered together. I would like to see this fixed, but seeing as this task is currently blocking Stapay, I am fine to call that a separate task and put that in the QA sprint

RyanASwanson (changes requested):
Part of the task description states that this can occur when walking through a trigger. I don't currently see functionality in place for that? I am fine to be flexible with some of my suggestions here purely as this task is a blocker right now, but as some aspects of the functionality are incomplete I am marking this as request changes. @ me as soon as changes occur.

nrice25 (approved):
I will let Ryan's request hold the PR, I don't have any other asks than what he has. 
In game it works as you said it would and bug fixing this issue would be a polish thing for later.

PR #530, opened 2025-02-27
beta-art-add-boat-asset-pack-to-engine

MilesRogers3D:
Added boat assets from asset store to the engine/updated materials to work with URP

Model file(s) location: **Assets/General/Art/ThirdParty/Warehouse Props Pack/Meshes**

Materials are already assigned, all you have to do is drag/drop the mesh into a scene


JamesBondMeow:
All meshes implemented in engine from asset pack and have textures

angelkait:
pop off king

ASLH68 (approved):
Approved on Micah's behalf

PR #536, opened 2025-02-28
Lv 911 change the esc back asset

hhlaxbro:
In the options menu, settings, controls, and notes, added art assets for the ESC: Back, and for notes, the A and D for when that functionality to swap between pages is implemented.

Jira: https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=61f6e440c6bd1a00691251d4&selectedIssue=LV-911


MagicTurtle013 (approved):
I like the placements, feedback listed in the PR thread

hhlaxbro:
> Changes seem alright to me. My only concern is that in terms of arrow placements within the note UI, there's no clear feedback to hovering over an arrow using the mouse to indicate that the arrows are clickable in addition to (for the future) browsable using key or button presses. I'd suggest, if there's time to do so, making a hover state for arrows to bring clarity to this.

The note UI itself was not touched. That is a separate task that I will handle next

PR #537, opened 2025-02-28
beta-des-Monster-Prefab-Refinement-LV-1250

JamesBondMeow:
- Changed a lot of values within the prefab to be more consistent
- Existing stem attack prefabs in scenes will most likely need to be adjusted according to the new base prefab values


BreadReaver (approved):
Looks good monster will for sure need to be moved but I think theyre gonna be taken out of the scene anyway so it works out

PR #539, opened 2025-02-28
fix-shadergraph-error

MilesRogers3D:
Edited the CelShaderStandard file to not use a custom GUI class. Builds on my machine(tm)


astapay (approved):
Yippee!

PR #542, opened 2025-03-01
Replaced old generator with new one

SymphonicSweet:
I just changed the prefab the code and colliders were on so that the correct generator was in place


pmetcalf2 (approved):
Generator is in, has needed functionality, go ahead and merge in.

PR #545, opened 2025-03-01
Beta eng brightness setting GitHub weirdness

JeremiahPeters311:
extra branch I created because github was being funky. This is for task https://bradleycapstone.atlassian.net/browse/LV-1147

adds the pre-existing post processing object to the universal managers prefab and creates a slider to modify the brightness (aka post exposure) inside it. 

known issue: the slider itself does not save its past position


Assets/General/Scripts/UI/VideoSettingsBehaviour.cs
```diff
@@ -0,0 +1,45 @@
...
+using UnityEngine.UI;
+using UnityEngine.Rendering;
+
+/// <summary>
+/// operates video settings, currently just brightness but probably more to come
+/// </summary>
+public class VideoSettingsBehaviour : MonoBehaviour
+{
+    private VolumeProfile volumeProfile;
+    private UnityEngine.Rendering.Universal.ColorAdjustments colorAdjustmentsName;
```
RyanASwanson, line 21: I feel this could be simplified with a using statement? Also these private variables should start with an underscore _

Assets/General/Scripts/UI/VideoSettingsBehaviour.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    [Tooltip("Should not be higher than like 5 or some visuals kinda break")]
```
RyanASwanson, line 23: You can use [Range(X,Y)] to limit the amount a float can be changed. (Obviously switch out X and Y for what they should be. Probably 0 - 5?)

Assets/General/Scripts/UI/VideoSettingsBehaviour.cs
```diff
@@ -0,0 +1,45 @@
...
+        if (!volumeProfile.TryGet(out colorAdjustmentsName)) throw new System.NullReferenceException(nameof(colorAdjustmentsName));
+    }
+
+    /// <summary>
+    /// change brightness value to match slider
+    /// </summary>
+    public void ChangeBrightness(Slider BrightnessSlider)
+    {
+        colorAdjustmentsName.postExposure.Override(BrightnessSlider.value * _maximumBrightness);
+        //needs something with save data to keep slider position
```
RyanASwanson, line 43: You should be able to mostly copy how the AudioSettings saved the data I would imagine

Assets/General/Scripts/UI/VideoSettingsBehaviour.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    /// <summary>
+    /// set up references
+    /// </summary>
+    private void Awake()
+    {
+        volumeProfile = GameObject.Find("GlobalVolumePostProcessing").GetComponent<UnityEngine.Rendering.Volume>()?.profile;
+
+        //even though this line is just an error check, everything breaks without it.
+        if (!volumeProfile.TryGet(out colorAdjustmentsName)) throw new System.NullReferenceException(nameof(colorAdjustmentsName));
```
RyanASwanson, line 34: Coding standards states this should be broken up into multiple lines

RyanASwanson (approved):
I feel mixed on whether to approve or request changes. I feel the brightness slider should save, but considering you still have a lot on your plate, I will be generous and approve with the caveat that the recommendations get addressed

nrice25 (approved):
Approved, the saving functionality is for another PR (we just need to make a jira ticket for it)

PR #546, opened 2025-03-01
stem(attack) prefab updated with hidden attack functionality

troberts2:
literally just changed the prefab of the attack to what was in my test scene


RyanASwanson (approved):
Seems fine to me

PR #547, opened 2025-03-01
Beta eng iterate on chase sequence lv 1139

troberts2:
You can run through the maze with the attacks currently in scene "ProceduralTest". The graphics are bugged and the textures are missing in the floor and walls so warning the level does look weird.


BreadReaver (approved):
Figure out who needs to look at textures at somepoint

pmetcalf2 (approved):
chase sequence is in and activates in all 3 straight aways

PR #553, opened 2025-03-02
add target decal

noodlecat519:
Added new decal texture and material in GameAssets/World/WorldArt/WallDecals. The decal is also on the wall in the VFXWork scene in General/Art/ArtTestingGyms. If one of these file pathways isnt correct lmk and I'll find it for you


hhlaxbro (approved):
Looks Good, approved

PR #558, opened 2025-03-02
made new prechase prefab based on Caden sketch

BreadReaver:
Made the transition area between maze and chase


pmetcalf2 (approved):
Wow, what a maze, update with main and go ahead and merge.

PR #559, opened 2025-03-02
beta-eng-Code-Functionality-for-Healing-SFX-LV-1325

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1325

There is now functionality for a healing sfx to play. I tested it with various sfx and it works. I removed the sfx before merging as we currently don't have any sfx for it in engine.

Code Estimation Time: <2
In Engine Estimation Time: <3


nrice25 (approved):
Approved, there is a slight bug that I ran into where the player didn't actually heal, but I think that was because of a bug?
If anything, something to note down and come back to whatever changes were made in this branch if that's the case moving forwards

astapay (approved):
Yippee!

PR #562, opened 2025-03-02
Monster Level Pass LV-1343

pmetcalf2:
Monster attacks are set up in a prefab in Narrative folder under prefabs, theres a line up cube for the water anomliy that you can use to get the monsters attacks set up in maze.


BreadReaver (approved):
Just make sure we cant goomba stomp them at some point in QA

PR #564, opened 2025-03-02
Beta eng hud inventory functionality lv 1295

JeremiahPeters311:
adds player hud ui showing the player's currently held item, in the form of an image with a fade in animation. In order to have an item trigger the hud, it needs to call setInventoryImage and pass in an int corresponding to the item (0 is key, 1 is spanner, 2 is gas can). In order to get rid of this when the item is used, disableInventoryImages should be called. I have this set up for the key and hatch in the game scene, but the key is currently broken so testing is much more complicated. 

(!!!) design will need to add the above mentioned functions onto any objects which give you a held item or use a held item (!!!)

https://bradleycapstone.atlassian.net/browse/LV-1295

also in this PR I quickly fixed a bug with the ammo racks interactable pop up not disappearing when out of harpoons, it was like two clicks in the prefab, idk how this got broken anyway.


RyanASwanson (approved):
Yeah looks good to me

nrice25 (approved):
It works, approved

PR #567, opened 2025-03-03
Beta UI update notes UI lv 1291

hhlaxbro:
Jira: https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=61f6e440c6bd1a00691251d4&selectedIssue=LV-1291

Updated the notes prefab when on open the UI will be more cohesive and clean.

Encountered a bug where the arrows to the right and left will not disappear when on the first and last page respectively.


Assets/GameAssets/World/WorldScripts/Interactables/NoteInteractable.cs
nrice25, line 1: I have no idea why there is code, but I think that you told me that Ryan did this so I trust it; it only seems to be a rotational thing anyways

nrice25 (approved):
Approved, it looks like it works in engine

PR #569, opened 2025-03-03
Lv 1353 chase build up asset implementation

Dusk553:
Added props to the generator room for the chase buildup prefab. If you could check to see they are there, it is in the duplicate called betachasebuildup (Props). Just checking to see if it looks fine and that everything is there is good.

Jura task: https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=63c32962c52a63dcda83f747&selectedIssue=LV-1353


MagicTurtle013 (approved):
remember to update with main

PR #574, opened 2025-03-03
beta-eng-Code-Functionality-for-Dropper-Water-SFX-LV-1323

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1323

Provides functionality to randomly play vfx. Also specifically for the water drop. I made a test scene called WaterDripTest. There are vfx over the stairs. The water drop is so incredibly subtle so I added an extra very obvious and very purple effect that also does the same thing. The audio I put in there is not 3D so it can be heard anywhere. This will change from audio in the future


astapay (approved):
Yippee!

PR #581, opened 2025-03-04
Updated Sound Tube Prefab MAT

angelkait:
Just some extra shading done in the outside tubes!


pmetcalf2 (approved):
The openings on the side tubes have darker shading in them, go ahead and merge

PR #588, opened 2025-03-04
Added specialized audio

Scottish02:
Specialized any SFX that might need it. 
Also organized all audio into folders.


MagicTurtle013 (approved):
i can see the sfx organization and that a number of them are spatialized

PR #593, opened 2025-03-04
Beta des hook up chase sequence lv 1385

BreadReaver:
Hooked up note found in chasebuildup so it correctly triggers chase, thank you Ryan


pmetcalf2 (approved):
map shifts now to the chase sequence at the end, go ahead and merge

PR #599, opened 2025-03-05
Beta eng hud objective functionality lv 1294

JeremiahPeters311:
Objective functionality is here at last! this includes both the pop up in the hud, as well as in the pause menu. As always, the way things look now is just a rough mock up by me and it will need to be designed properly to look better by Hayden probably.

In order for anything to trigger an objective change, it needs to trigger ActivateObjectiveHud for the hud and SetPauseMenuObjective for the pause menu. I set this up for the key in GameScene to use as an example and for testing, but any other objectives will need to be setup by design.

https://bradleycapstone.atlassian.net/browse/LV-1294
should be a relatively quick review for the code part, but will need some tweaking for the visuals as mentioned.


RyanASwanson (approved):
Just some notes about adhering to the naming conventions. The functionality looks good. Could certainly use some cleanup for UI/UX but that falls on Hayden and Adam. Good work on this, it works quite seamlessly!

hhlaxbro (approved):
Baller! Works great!

astapay (approved):
Hooray!

PR #603, opened 2025-03-05
beta-eng-Implement-2D-Cinematics-LV-1270

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1270

Implemented the cinematic files and made the scene for the outro. The outro is now in the build settings and after it plays it goes back to the main menu. I can't currently hook up the hatch to go to specifically the outro scene as the maze is constantly being changed, but its as simple as design just swapping an integer. I also implemented the functionality for the audio to play at the start of cinematics. Currently I am just using the music as test audio for these as we don't currently have the full dialogue in engine yet. (You may notice the music continuing after the cinematic but that is just because they are longer than the cinematic, this is simply resolved by having the correct audio).

You can test this by playing the title screen and pressing play, it will go to the intro cinematic, then afterwards goes to gameplay. As mentioned above the outro can currently only be accessed by going to the scene directly. It goes to the main menu after completion.

Code review estimation: <3 minutes

### DIFF
--- a/Assets/General/Audio/FMODScripts/FmodSfxEvents.cs
+++ b/Assets/General/Audio/FMODScripts/FmodSfxEvents.cs
@@ -108,6 +108,11 @@ public class FmodSfxEvents : MonoBehaviour
     [field: SerializeField] public EventReference WaterDrip { get; private set; }
     #endregion
 
+    #region Cinematics
+    [field: Header("Cinematics")]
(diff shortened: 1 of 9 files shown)


nrice25 (approved):
Approved, but I would highly recommend adding the functionality for audio to stop after the cut scene is over

astapay (approved):
Yippee!

PR #605, opened 2025-03-05
Beta des implement temporary engine room voiced dialogue lv 1123

pmetcalf2:
VA lines for dialogue in the story beat system have been implemented. Let me know if the timing seems off between the text coming on screen and the voices (Player lines come out slow; I can't make them go out any faster).


BreadReaver (approved):
Text speed could maybe be changed but thats semantics

PR #616, opened 2025-03-06
beta-eng-Lights-change-intensity-after-flicker-animation-plays-LV-1233

astapay:
Wow i did the thing! The lights should be a tad more consistent. After they flicker, they will be set to the same intensity. I could change it such that it is relative to the original intensity, but that would require an entirely new script so if we want that it'll be a longer task.

Also Oliver requested that I made brightness a bit stronger so I did that too :D


RyanASwanson (approved):
Yup it remains at an intensity of 5

ogNINJAGOLDBOX (approved):

PR #619, opened 2025-03-06
beta-des-Music-Implementation-LV-1372

ogNINJAGOLDBOX:
Put in the Title Music and Death Screen Music.

Put in triggers on the monster and a bit of code (Ryan directed me) to play the Monster Music when within range of the spawn.

https://bradleycapstone.atlassian.net/browse/LV-1372

The press play in the Title and Death screen and those should automatically play

There is a test gym called MusicTestGym2 where there is a monster and you can test the swapping between Monster Music and Waves Ambience


astapay (approved):
Yippee!

PR #621, opened 2025-03-06
Additional Controls Menu Page LV-1146

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-1146

adds the the controller controls to the controls menu and functionality to switch between keyboard and controller pages. code is simple and basically copied from notes. should be a quick review.


Assets/General/Scripts/UI/SwapControlsPage.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            _keyboard.SetActive(false);
+            _controller.SetActive(true);
+            _leftButton.color = Color.clear;
+            _rightButton.color = Color.white;
+        }
+        else if (value == -1)
+        {
+            _keyboard.SetActive(true);
+            _controller.SetActive(false);
```
RyanASwanson, line 49: Technically we could write this shorter by saying _keyboard.SetActive(value== -1) and have that outside the if statement. This isn't something I am requiring you to change, just something to think about

RyanASwanson (changes requested):
There is a mismatch on the controls screen for the arrows not being the same direction as the input. Just address that and I will look again and most likely approve

RyanASwanson (approved):
Yup. He swapped the input

nrice25 (approved):
It works in game now, approved

hhlaxbro (approved):
Works. nice nice

PR #624, opened 2025-03-06
Lv 1298 hud and pause objective implementation

hhlaxbro:
Jira: https://bradleycapstone.atlassian.net/browse/LV-1298?atlOrigin=eyJpIjoiNTc4ZjYyMGJiMTQ0NDM4NTg4Mjk3YmJlMGM0MDk1YTYiLCJwIjoiaiJ9

Slight visual change to the objective pop-up and in the one in the pause. Known issue that the pop-up text is too big and will not change when altering the text size in the HUD.


nrice25 (approved):
Visually this is an improvement; albeit, the bug with the pause menu not changing objective should be turned into a jira ticket.

PR #628, opened 2025-03-08
Title Screen Apparent Fix

SolarFoxG:
Made the background of the title screen's camera into a black BG instead of a skybox since it can become visible when the player's device screen is wide. Less obtrusive than skybox.

Test this by putting the game into Free Aspect. On most displays, this should never be noticed, but my computer specifically has a wide enough screen to where the menu art isn't wide enough to scale in full, and other players may reasonably have the same issue.


hhlaxbro (approved):
Looks nice! things are scaled weirdly in free aspect mode, text not being on screen, things being cut off, but that does feel like a bigger problem. I'm not sure if you want to tackle that but I'll approve for now.

PR #630, opened 2025-03-08
Beta UI adjust settings layout lv 1283

SolarFoxG:
Jira Issue: https://bradleycapstone.atlassian.net/browse/LV-1283

Cleaned up the layout of the settings submenu, and also ensured that settings scale with different screen resolutions without awkwardly stacking or obscuring any options or information.


hhlaxbro (approved):
Works great!

PR #631, opened 2025-03-08
Beta UI cinematic skip button implementation lv 1290

hhlaxbro:
Jira: https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=61f6e440c6bd1a00691251d4&selectedIssue=LV-1290

Super light change giving the skip a BG as well as changing the font and moving it. Ideally this version of the skip version will just be temporary. I think in gold it's functionality needs to change to be more refined.


nrice25 (approved):
Approved, looks good

PR #632, opened 2025-03-08
beta-eng-Update-player-start-rotation-LV-1378

astapay:
Player start rotations updated and can now be set via the Player Spawn Point. Just set the rotation of that game object and the player will spawn with that rotation.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-1378?atlOrigin=eyJpIjoiMmVlOGIzZDc5ZDg1NDMyZjliNTk4YmQyMDE1YjkwNjIiLCJwIjoiaiJ9

Code Review Time: <2 minutes
In-Engine Review Time: <5 minutes


RyanASwanson (approved):
Yeah the player spawns facing the correct direction now. Just curious why this was a change in the scene and not the prefab?

nrice25 (approved):
Works in game, I don't really understand the math but I don't need to
Approved

PR #633, opened 2025-03-08
beta-des-Voice-Tube-Duplication-LV-1429

JamesBondMeow:
- After the player grabs the spanner the voice tube is duplicated and clips into itself (fixed)


BreadReaver (approved):
Duplicate successfully deleted

PR #634, opened 2025-03-08
beta-des-Stretched-Health-Pack-LV-1430

JamesBondMeow:
- Adjusted health pack prefab so its transformation is no longer stretched


BreadReaver (approved):
Health pack sizes are suitably normalized

PR #635, opened 2025-03-08
beta-des-2nd-Prop-Pass-for-Normal-Level-LV-1444

JamesBondMeow:
- Added 63 more props to the normal level prefab to make it feel more cluttered


BreadReaver (approved):
Feel like stuff could be more scattered as opposed to just pressed against walls but we can talk about that in the future

PR #638, opened 2025-03-08
beta-eng-Realign-Interaction-and-popup-LV-1388

Titonfish:
https://bradleycapstone.atlassian.net/browse/LV-1388?atlOrigin=eyJpIjoiMzNkYWQyMjgyZDA0NGY1NWJkNzY0MmJlMWVmMGM1MTMiLCJwIjoiaiJ9
Added the ability to check what the player was currently looking at, and made sure that the popup only worked if they could interact with it


Assets/GameAssets/UI/UIScripts/WorldSpace Popups.cs
```diff
@@ -98,6 +114,7 @@ private IEnumerator FindPlayer()
         {
             _playerCamera = pfc.PlayerCamera.transform.Find("Main Camera").GetComponent<Camera>();
             _playerReference = pfc.transform.GetChild(1).gameObject;
+            _playerInteractor = pfc.GetComponentInChildren<PlayerInteraction>();
```
RyanASwanson, line 117: Maybe at this point we should just Instance the PlayerInteraction?

RyanASwanson (approved):
Approved, just left a note about using the GetComponentInChildren for each interactable.

PR #639, opened 2025-03-08
Placeholder Tutorial Functionality LV-1404

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-1404

inside TempTutorialTestGym scene you can find a red cube that when you get near, opens an extremely note-like tutorial screen with multiple pages that can be controlled just like a note and exited with escape. This object is also a prefab called PlaceholderTutorialTrigger which has no red cube but otherwise is exactly the same

almost all of the code in this is copy pasted from notes so unless i messed up a name or forgot a comment there should be no issues in review.


astapay (approved):
Yippee!

RyanASwanson (approved):
Sorry for the delay. This looks good to me. I'm generally not too picky about Placeholder stuff anyways.

PR #642, opened 2025-03-08
beta-eng-Settings-sliders-stay-from-multiple-playthroughs-while-settings-reset-LV-1411

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1411

So this issue wasn't entirely what the description says on Jira. The issue was that the audio volumes ONLY ever got set when you adjusted the sliders. So if you loaded the game, the save data was actually loaded properly (despite what the Jira ticket says), its just that the volumes were never changes.
Easiest way to test this is to go into the game, and set the music volume to 0. Then reopen the game and you will now not hear any music on the main menu. 
I initially thought the problem was something else, and in doing so I identified and fixed a seperate issue as well. The slider settings only changes when you dragged the settings. If you just clicked to move the settings it never actually saved that. (I thought that was the cause of the issue, turns out it wasn't, but I fixed it anyway 🤷 


astapay (approved):
Yippee!

PR #644, opened 2025-03-09
Fix Popup Rotation LV-1387

JeremiahPeters311:
very simple change, just made it so the pop ups rotate up and down not just side to side

https://bradleycapstone.atlassian.net/browse/LV-1387


astapay (approved):
Yippee!

PR #645, opened 2025-03-09
Beta des noticeable map shifts lv 1438

pmetcalf2:
All the generator rooms should have all the same props and be less noticeable when the shifts happen. (All prefabs except chase sequence)


BreadReaver (approved):
Nice work

PR #646, opened 2025-03-09
beta eng Heart Animation Bug Repeated fade ins and outs LV 1026

Titonfish:
https://bradleycapstone.atlassian.net/browse/LV-1026?atlOrigin=eyJpIjoiM2MzMWJlY2JhYzVhNDViZjhlOWZlZmJhZGJlOTBiZWEiLCJwIjoiaiJ9
Changed the heart visibility coroutine so that it takes into account the current visibility


RyanASwanson (approved):
Works in all situations it seems. I tried taking damage while the ui was fading in, was currently present, and was fading out and it appeared to be working properly.

PR #652, opened 2025-03-09
Camera velocity is limited on mouse and keyboard lv 1218

C-Low01:
Camera velocity stuff changed, max speed is higher, startup time changed from half a second to a second for more fluidity, and camera runs on fixed update now instead of late update.


TurquoiseDragon:
Everything seems to work as expected!

ogNINJAGOLDBOX (approved):

MagicTurtle013 (approved):

PR #654, opened 2025-03-09
Monster Attack Animation Breaks LV 1451

pmetcalf2:
The 2 monsters that spawned weird due to prop collision and spanner moved back to where it should be.


BreadReaver (approved):
Monsters dont appear to get caught on anything

PR #658, opened 2025-03-10
LV-1360-Spark-VFX-doesn-t-appear-on-wall

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1360

Updated the harpoon vfx check to also take into consideration all colliders and not just mesh renderers. There is a test scene I used called WeaponVFX testing, in which I just shot random stuff (including the floating cube) in the environment to see if the vfx were playing properly. You can shoot various objects and they either play spark or wood vfx depending on what is hit. I updated the materials that the game checks. It doesn't include every material, I just added like 6-7, but that's a design thing. The vfx are rather subtle, so its hard to notice when the spark/wood vfx plays, but they are there. Especially the wood vfx is hard to see. I made it larger, but still.

In Engine Review Estimation: <8 minutes.


nrice25 (approved):
Approved, works in engine and the code has been improved to allow regular colliders

PR #662, opened 2025-03-10
Pause-death-bug-is-back-LV-1467

nrice25:
- The tutorial will no longer prevent the mouse from being seen across UI screens
Fixed the tutorial bug I also brought the note interactable to parity and fixed errors from popping up from that script


TurquoiseDragon:
Tested it, everything seems all up to snuff

RyanASwanson (approved):
I like how this is cleaned up to use the time manager. Much better

PR #667, opened 2025-03-10
beta-des-Reduce-monsters-in-hallway-LV-1383

ogNINJAGOLDBOX:
Tried to lower the amount of monsters without losing the idea of having them grow more and more frequent

https://bradleycapstone.atlassian.net/browse/LV-1383

Just play through the Maze Scene, Beta Maze and see how you feel about the frequency of the monsters. You may also have to look in the unity scene as some of them still spawn weird so it is hard to see them


BreadReaver (approved):
Looks good

PR #668, opened 2025-03-10
Beta eng paused audio lv 1421

Titonfish:
https://bradleycapstone.atlassian.net/browse/LV-1421?atlOrigin=eyJpIjoiNTdkNjRhNWI2M2JmNGFiNGI5NWU2MTgxYzViM2I1ZDgiLCJwIjoiaiJ9
Made the audio mute on pause, essentially a poor man's pause audio function


RyanASwanson (approved):
Yeah this is one I gotta tentatively approve of. I get that you don't really like this solution, and neither do I, but with 3 hours left in the milestone I cannot expect a full audio pause to work. Not really much more can be done about it so I will approve it.

PR #672, opened 2025-03-10
Lv 1367 fix pop up placement and size

hhlaxbro:
Jira: https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=61f6e440c6bd1a00691251d4&customFilter=ari%3Acloud%3Ajira-software%3A792ff6f6-f5ba-4734-a1cd-457d34518a06%3Acustom-filter%2Factivation%2F489b3b11-a1a5-4c9e-afed-c8d621aa5e68%2F32%2Fe7f130a4-1698-4dfa-a009-a1806a05600f&selectedIssue=LV-1367

Updated the key, notes, and health pack to fix the pop-up stretching. Now the scale of all of those should be set to 1, meaning not strange stretching. Since play is currently broken, best way to check is to go into the hierarchy of each one of these and if the scale is one it should work.


nrice25 (approved):
Approved

PR #674, opened 2025-03-11
beta eng Add Audio to Cinematics LV 1476

Titonfish:
https://bradleycapstone.atlassian.net/browse/LV-1476?atlOrigin=eyJpIjoiZWU0MzVjMjhmZjZhNDUzM2FlZGM1YzU4NzRiOGZlMTAiLCJwIjoiaiJ9
Fixed the audio so that now it actually stops during the cutscenes, and doesn't carry it over to the title screen


RyanASwanson (approved):
Glad you figured this out. Still feels weird having the cinematic audio play differently than everything else, but if audio can't figure that out it's not on you.

PR #675, opened 2025-03-11
pop up and inventory fixed

JeremiahPeters311:
turned off generator pop up by default, added event to key to turn it on when picked up, as well as adding it to the inventory hud


hhlaxbro (approved):
Noice!

RyanASwanson (approved):
Looks good to me!

PR #679, opened 2025-03-11
Beta-Eng-Chase-SFX-LV-1474

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?selectedIssue=LV-1474

The chase audio now properly follows the monster. It just kinda played where the monster spawned and that was it. As you got further away you just couldn't hear it anymore. This is a bit of an awkward one to test but bare with me. A) Either play the game normally and get to the chase sequence or B) Go into the ChaseTesting scene and press play. This then loads in the first maze at start because that's how Mile's system works. This creates an overlap between the chase sequence and the first maze. Just disable the first maze by disabling Maze_1_Root. Then you are now just in the chase sequence. 
Go to the end of the dark hallway and the chase sequence starts up as normal. You can hear the moving audio playing properly.

In Engine Review Estimation: <5-10 minutes depending on testing method


astapay (approved):
Yippee!

PR #680, opened 2025-03-11
Mouse Disappears after Tutorial Bug LV-1475

JeremiahPeters311:
not sure if this will fix the issue but the behavior should be the same as for the notes now


RyanASwanson (approved):
Yup. Good work both Jeremiah and Charlie.

PR #683, opened 2025-03-12
Beta eng open pause and settings loads maze lv 1477

nrice25:
The scene should no longer load in the next maze on top of the current maze when going to another subscene
IE: Settings or Quit

There might be errors beyond the first maze, but I'm exhausted and don't have the brain power to be on my laptop


RyanASwanson (approved):
This is all good!

PR #692, opened 2025-03-26
gold-art-fix-camera-sensitivity-LV-1530

MilesRogers3D:
Made some general improvements to the camera sensitivity and Cinemachine camera movement. The harpoon gun is still a little jittery but I think that's because of the fact that everything is running on coroutines, which probably means there's a race condition for writing the position/rotation of the gun. I was able to massively improve this issue by decoupling it it from the camera, though

Testing notes:
- I do not have a controller to test with currently, so if someone could confirm I haven't somehow broken controller support that would be amazing


RyanASwanson (changes requested):
The harpoon following is bugged. It is supposed to stay locked in the center of the screen while aiming down sights.

RyanASwanson (approved):
Still fix that XML, but you justified my other note

PR #695, opened 2025-03-26
Gold UI inventory icons polish lv 1496

SolarFoxG:
Nested inventory icons better in the bottom right corner, and also reduced their size. I'd like to note that icons still scale oddly based on resolution, but this is a larger issue with the player HUD as a whole that needs to be addressed.


nrice25 (approved):
Approved, if someone with a different aspect ratio says it isn't working we will come back to this; but from what I tested it's fine

PR #706, opened 2025-03-28
Gold eng subtitle settings lv 1518

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-1518

makes the subtitles option in video settings operational. this includes adding save support as well as a slight bit of code to DialoguePopUps.cs to actually make the setting do as it suggests.


astapay (approved):
Yippee!

hhlaxbro (changes requested):
I think that it needs to be tweaked where if subtitles are disabled, then they turn it back on mid, dialogue, the subtitles should turn back on, not wait till the next line of dialogue for the subtitles turn on.

hhlaxbro (approved):
Works Great Approved

RyanASwanson (approved):
All good, just left a single note

PR #707, opened 2025-03-28
gold-eng-fix-lights-prefab

astapay:
Now we return colors yay!

In-Engine Review Time: <5 minutes (whenever you get to the point where the lights change color)


RyanASwanson (approved):
Yippee. Sorry for the delay

PR #708, opened 2025-03-29
gold-des-Fog-Pass-LV-1559

ogNINJAGOLDBOX:
Changed the fog in the scenes:
- In maze scene it comes and goes throughout by rising and falling
- In chase sequence is rises higher and higher as you get to the exit

https://bradleycapstone.atlassian.net/browse/LV-1559

All you need to do is play through the mazes as normal. The red lights might make it hard to see, those won't stay red forever though, so if you want to see what exactly I did just look in editor


Dusk553 (approved):
I approve under the stipulation for the change that I put in the thread, just so that when you get approved by kyle after such changes, you aren't left dead in the water waiting for longer than nessesary

PR #709, opened 2025-03-29
gold-eng-Normalized-Flicker-Animations-LV-1719

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1719
This shouldn't actually feel different from main at all. I reworked the way the light flicker functions from being on an animation to being done through script. The reason is that you can modify the intensity of the light and the flicker will scale based on that. (It doesn't update at runtime. It sets its starting intensity to scale off of in start)
Honestly making the animation curves match was what took the longest.

Feel free to modify the intensity of a light in maze 2 then try playing the game (from the maze scene) and that lights flicker will scale properly.


nrice25 (approved):
Approved, the only errors I noticed were not because of your branch

astapay (approved):
Yippee!

PR #713, opened 2025-03-30
Gold art wall crack decal lv 1552

noodlecat519:
Added 4 new decals (monster eye, water stain, dirt smudge, crack) and updated the hole decal. New decals are on the wall in the first room of the first maze in the scene DecalMazeTesting. It is in Assets/Scenes/TestGyms/ArtTestGyms.


angelkait (approved):
everything looks good! great job caelie, you're killing it! <3

PR #717, opened 2025-03-30
gold-eng-Walking-Audio-DeathBug-LV-1712

Lazydev114:
Jira Link: https://bradleycapstone.atlassian.net/browse/LV-1712
Description: When you die, you can still hear the walking footsteps. had trouble replicating the bug perfectly, but I put in a fix, and there don't seem to be foot step sounds when I die. 
Best Tested: Play the game and try to die. If there is no footsteps, then it works

### DIFF
--- a/Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/RuntimeSfxManager.cs
+++ b/Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/RuntimeSfxManager.cs
@@ -31,6 +31,8 @@ public class RuntimeSfxManager : AudioManager
 
     private WaitForSeconds footstepDelay;
     private WaitForSeconds firstFootstepDelay;
+    //determins if the foot steps should be played or not
+    public bool CanPlayFootSteps = true;
 
     #region Enable and Action Subscriptions
     /// <summary>
@@ -281,6 +283,7 @@ public class RuntimeSfxManager : AudioManager
     /// </summary>
     private void PlayFootStep()
     {
+        if(CanPlayFootSteps == false) return;
         if (PlayerMovementController.IsGrounded && PlayerMovementController.IsMoving)
         {
             if (_currentWalkingSfx.IsNull)
@@ -299,6 +302,7 @@ public class RuntimeSfxManager : AudioManager
     /// <returns></returns>
     private IEnumerator LoopFootSteps()
     {
+        if (CanPlayFootSteps == false) yield return null;
         // Update the initial footstep speed
         float currentSpeedMultiplier = PlayerMovementController.Instance.CurrentFocusMoveSpeedMultiplier;
         firstFootstepDelay = new WaitForSeconds(FmodSfxEvents.Instance.FirstFootstepDelay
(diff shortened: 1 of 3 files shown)


nrice25 (approved):
Approved, it seems to fix the bug; the issue with not having a mouse in the death scene should added to the bug list

Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/RuntimeSfxManager.cs
```diff
@@ -281,6 +283,7 @@ private void StopFootsteps()
     /// </summary>
     private void PlayFootStep()
     {
+        if(CanPlayFootSteps == false) return;
```
RyanASwanson, line 286: I feel like this can be done in a way that doesn't require an if check every footstep.

Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/RuntimeSfxManager.cs
```diff
@@ -299,6 +302,7 @@ private void PlayFootStep()
     /// <returns></returns>
     private IEnumerator LoopFootSteps()
     {
+        if (CanPlayFootSteps == false) yield return null;
```
RyanASwanson, line 305: Why not just stop the coroutine in this case? Why have it yield return null every frame? Also for coding standards the yield return null should be in brackets in a separate line

Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/RuntimeSfxManager.cs
```diff
@@ -31,6 +31,8 @@ public class RuntimeSfxManager : AudioManager
 
     private WaitForSeconds footstepDelay;
     private WaitForSeconds firstFootstepDelay;
+    //determins if the foot steps should be played or not
+    public bool CanPlayFootSteps = true;
```
RyanASwanson, line 35: This is under a universal manager. Universal managers persist for as long as the game is running and is never destroyed. As such, the only time this boolean is set to true is on opening the project and it is never turned back on. So if you die and replay the game there would be no footsteps

RyanASwanson (changes requested):
I feel this system can be done in a way such that there doesn't need to be an if statement for every frame. Also there is an unintended bug of footsteps never being reenabled

Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/RuntimeSfxManager.cs
Lazydev114: Ah, ok, I see. I'll fix this asap

Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/RuntimeSfxManager.cs
```diff
@@ -221,6 +222,14 @@ private bool CheckForNullSfx(EventReference eventReference)
 
     #region Footsteps
 
+    private void OnLevelWasLoaded(int level)
```
RyanASwanson, line 225: We should probably just have a basic function subscribed to the event on the AframaxSceneManager for when a scene is loaded, just to remain consistent. Or we could simply have it reenable in the death screen.

RyanASwanson (approved):
Approved with a single note

PR #722, opened 2025-03-31
gold-eng-Rain-Patch-SFX-Functionality-LV-1711

Lazydev114:
Jira Link: https://bradleycapstone.atlassian.net/browse/LV-1711
Description: A script that plays a specific FMOD sound persistently. The rain sound is not in there yet, so I just used the death screen sound because it had loop enabled. The sound used needs to have loop enabled. This can only be done in fmode
Best tested: go to the persistent audio manager and find the persistent sound script. Change the slot to an fmode event that has a loop on. To enable loop, go into fmode and add a loop region


Assets/Scripts/Sound/PersistentSound.cs
```diff
@@ -17,17 +17,17 @@
...
+    [SerializeField] private int _soundId;
     /// <summary>
     /// happens whent he scene starts
     /// it triggers the fmode event
     /// </summary>
     private void Start()
     {
-        _eventInstance = RuntimeManager.CreateInstance(Sound);
+        EventReference[] sounds = FmodPersistentAudioEvents.Instance.PersistentSound;
+        _eventInstance = RuntimeManager.CreateInstance(sounds[_soundId]);
```
RyanASwanson, line 30: I don't believe we are ever setting the 3D location the audio plays.
The AudioManager have a function called
CreateInstanceFromReference which can take in a gameobject to play the sound on. That should probably be used instead.
Also a test gym would help to verify if it's working or not

Assets/Scripts/Sound/PersistentSound.cs
```diff
@@ -17,17 +17,17 @@
...
 {
-    [field: SerializeField]
-    public EventReference Sound;
+    
     private EventInstance _eventInstance;
-
+    [SerializeField] private int _soundId;
     /// <summary>
     /// happens whent he scene starts
     /// it triggers the fmode event
```
RyanASwanson, line 25: should say fmod not fmode

Assets/Scripts/Sound/PersistentSound.cs
```diff
@@ -17,17 +17,17 @@
...
     /// </summary>
     private void Start()
     {
-        _eventInstance = RuntimeManager.CreateInstance(Sound);
+        EventReference[] sounds = FmodPersistentAudioEvents.Instance.PersistentSound;
+        _eventInstance = RuntimeManager.CreateInstance(sounds[_soundId]);
         _eventInstance.getDescription(out EventDescription eventDesc);
         eventDesc.isOneshot(out bool isOneShot);
         if (isOneShot) Debug.LogWarning("The event is a one shot, go in fmode and add loop region");
```
RyanASwanson, line 33: Same comment about fmode

RyanASwanson (changes requested):
I don't believe the audio is being set in it's location in 3d space. That would probably help a lot. Aside from that left some minor comments about spelling

Assets/Scripts/Sound/PersistentSound.cs
Lazydev114: for testing, I have two sounds in the list, just switch between them and you should hear a different sound

Assets/Scripts/Sound/PersistentSound.cs
```diff
@@ -20,18 +20,22 @@ public class PersistentSound : AudioManager
...
-        EventReference[] sounds = FmodPersistentAudioEvents.Instance.PersistentSound;
-        _eventInstance = RuntimeManager.CreateInstance(sounds[_soundId]);
-        _eventInstance.getDescription(out EventDescription eventDesc);
-        eventDesc.isOneshot(out bool isOneShot);
-        if (isOneShot) Debug.LogWarning("The event is a one shot, go in fmode and add loop region");
-        _eventInstance.start();
+        if (_shouldPlay)
+        {
+            EventReference[] sounds = FmodPersistentAudioEvents.Instance.PersistentSound;
+            _eventInstance = PersistentAudioManager.Instance.CreateInstanceFromReference(sounds[_soundId]); ;
```
RyanASwanson, line 33: Look at the message I sent in discord about this, there are 2 versions of this function.

Assets/Scripts/Sound/PersistentSound.cs
```diff
@@ -20,18 +20,22 @@ public class PersistentSound : AudioManager
...
+            _eventInstance.getDescription(out EventDescription eventDesc);
+            eventDesc.isOneshot(out bool isOneShot);
+            if (isOneShot) Debug.LogWarning("The event is a one shot, go in fmod and add loop region");
```
RyanASwanson, line 36: We try to avoid keeping an entire if statement and result in one line. Seperating things with brackets helps with readability even if there is only 1 line as a result. So this should look like
if (isOneShot) 
{
Debug.LogWarning("The event is a one shot, go in fmod and add loop region");
}

RyanASwanson (changes requested):
Left a few notes. Sorry about the confusion for the first comment

RyanASwanson (approved):
Yup approved!

PR #723, opened 2025-03-31
Pause Menu Quit Revisions LV-1701

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-1701

I was not expecting this to require code at all but unfortunately it did, as just loading the title screen does not work. So, I made a new function in the scene manager that destroys all dont destroy on load objects in order to go back to the title screen properly. Also this replaces the quit option in the pause menu with title screen, but I can add back in the quit option if it is desired. (you can still quit from the title screen)


RyanASwanson (changes requested):
Follows code standards well, with some minor suggestions but the issue of everything universal being destroyed leads to issues with tweens not working as they should.

hhlaxbro (approved):
Do whatever Ryan says but to me works good!

RyanASwanson (approved):
All good! Should be good to merge now!

PR #725, opened 2025-03-31
Gold des re mazeing the maze lv 1537

BreadReaver:
Added 3 new alternate hallways for the player to go down in order to give them a little bit more agency and the illusion of choice while they are exploring


TurquoiseDragon (approved):
Everything seems to be in order, No more collision or Z fighting issues that I could see

PR #739, opened 2025-04-02
Gold des decal pass lv 1577

Dusk553:
I added in several decals and was wondering your thoughts on them. I noticed that several of them don't look as good while the lights aren't on them and was wondering if there was any way to change that. If nothing else, I think this works as a first pass, considering that I don't know if there are going to be any more changes to the maze.

Jura task: https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=63c32962c52a63dcda83f747&selectedIssue=LV-1577


BreadReaver (approved):
Loving all the decal work, really bumps the environment

PR #745, opened 2025-04-03
gold-art-datamoshing-effect-lv-1527

MilesRogers3D:
Implemented a datamoshing effect for anomaly events.

image

List of everything I did code-wise:
- Added 2 new scripts, DatamoshEffectManager.cs and TriggerDatamoshingEffect.cs
- Added in a library called [URPGrabPass](https://github.com/Haruma-K/URPGrabPass)
- GitHub may have tracked a few changes to other script files that I did during testing, but they were all reverted to the way they were before

How to test:
- The effect has been implemented in the second maze, when rounding the corner after the first monster jumpscare (the one with the hand carts). When entering the volume in the scene it will trigger the effect for 1 second in a singleton class attached to the camera


Assets/Resources/SaveData/Data.json
astapay, line 1: I recommend removing the changes to this file so that we have a fresh save for future playtesting purposes

Assets/Scripts/HorrorMoments/TriggerDatamoshingEffect.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <summary>
+    /// Trigger the datamoshing effect when the player enters
+    /// the object's BoxCollider trigger
+    /// </summary>
+    /// <param name="other">Other collider</param>
+    private void OnTriggerEnter(Collider other)
+    {
+        if (other.gameObject.CompareTag("Player"))
+        {
+            GetComponent<BoxCollider>().enabled = false;
```
astapay, line 47: You may want to save the BoxCollider in a variable for efficiency.

astapay (approved):
Yippee!

Assets/Scripts/HorrorMoments/TriggerDatamoshingEffect.cs
MilesRogers3D: Good catch, fixed this!

RyanASwanson (approved):
No issues with the code, all good on that front. Can't really test the visuals obviously. I do worry that players will see the effect and assume it's a bug tho

PR #747, opened 2025-04-03
gold-art-reduce-project-file-size-lv-1531

MilesRogers3D:
Got the project size down to under a gigabyte in MinSizeRel build mode by tweaking some texture settings. This also improves the performance by quite a bit


RyanASwanson (approved):
Yup. This looks good to me!

PR #759, opened 2025-04-05
gold-eng-Monster-Encounter-Music-does-not-play-LV-1743

Lazydev114:
Jira Link: https://bradleycapstone.atlassian.net/browse/LV-1743of 
Description: The monster encounter music plays now. There were layers of crazy, messed-up scripts. I would have had to rewrite a lot of the code, so I just made a simple trigger that can start a song on start or stop. I don't know why this functionality was so overcomplicated. Also, I know this wasn't the task, but I tried finding out why the monster encounter music plays in the outro and how the hell it was doing that without the persistent sound manager, but I couldn't. I think we have a ghost in our game, guys.


Assets/Scripts/Sound/MusicTrigger.cs
```diff
@@ -0,0 +1,86 @@
...
+
+
+    /// <summary>
+    /// happens every frame
+    /// being used to make the object follow the player
+    /// </summary>
+    private void Update()
+    {
+       
+    }
```
RyanASwanson, line 32: Delete this if it isn't being used

Assets/Scripts/Sound/MusicTrigger.cs
```diff
@@ -0,0 +1,86 @@
...
+/// <summary>
+/// plays a sound on enter or exit
+/// </summary>
+public class MusicTrigger : MonoBehaviour
+{
+    private EventInstance _eventInstance;
+    [SerializeField] private int _soundId;
+    [SerializeField] private bool _playOnEnter = false;
+    [SerializeField] private bool _playOnExit = false;
+    private bool _playDoOnce = true;
```
RyanASwanson, line 22: booleans start with words such as do, does, can, is, etc.

Assets/Scripts/Sound/MusicTrigger.cs
```diff
@@ -0,0 +1,86 @@
...
+    private void OnTriggerEnter(Collider other)
+    {
+        
+        if (other.gameObject.tag == "Player" && _playOnEnter && _playDoOnce)
+        {
+            EventReference[] sounds = FmodPersistentAudioEvents.Instance.MusicInGame;
+            _eventInstance = PersistentAudioManager.Instance.CreateInstanceFromReference(sounds[_soundId]);
+
+            RuntimeManager.AttachInstanceToGameObject(_eventInstance,
+                GameObject.FindGameObjectWithTag("Player").transform ,
```
RyanASwanson, line 47: Can we use Instanced scripts for stuff like this. FindGameObjectWithTag is wildly inefficient. Also since this is music, and it is 2d. It doesn't need to be attached

Assets/Scripts/Sound/MusicTrigger.cs
```diff
@@ -0,0 +1,86 @@
...
+using JetBrains.Annotations;
+using UnityEngine;
+
+/// <summary>
+/// plays a sound on enter or exit
+/// </summary>
+public class MusicTrigger : MonoBehaviour
+{
+    private EventInstance _eventInstance;
+    [SerializeField] private int _soundId;
```
RyanASwanson, line 19: Maybe this could have a sound for entering and one for exiting? Just could be more versatile that way.

Assets/Scripts/Sound/MusicTrigger.cs
```diff
@@ -0,0 +1,86 @@
+/******************************************************************************
+// File Name:       MusicTrigger.cs
```
RyanASwanson, line 2: We already essentially have a script that does this exact thing. I mean I don't think it has the function to play music on leaving the area, but that's it. The script is called MusicSwapPlayerTrigger

RyanASwanson (changes requested):
Requires some revisions to clean things up. Also I am confused as to why a new script was made to do the same thing as MusicSwapPlayerTrigger.cs

Assets/Resources/SaveData/Data.json
astapay, line 1: You may want to remove the changes made to this file for future playtesting purposes. Just to make sure we don't already have a save in the game.

Assets/Scripts/Sound/MusicTrigger.cs
```diff
@@ -0,0 +1,86 @@
...
+       
+    }
+
+    /// <summary>
+    /// plays a sound on enter
+    /// </summary>
+    /// <param name="other"></param>
+    private void OnTriggerEnter(Collider other)
+    {
+
```
astapay, line 40: (Optional) may want to remove the whitespace here

Assets/Scripts/Sound/MusicTrigger.cs
```diff
@@ -0,0 +1,86 @@
...
+            _playDoOnce = false;
+        }
+    }
+    /// <summary>
+    /// plays a sound on exit
+    /// </summary>
+    /// <param name="other"></param>
+    private void OnTriggerExit(Collider other)
+    {
+
```
astapay, line 60: (Optional) and here

astapay (approved):
Get Ryan's things done and we should be good

Assets/Scripts/Sound/MusicTrigger.cs
FantasticGames1: i felt like the 3d effect of the sound following the player sounded better

Assets/Scripts/Sound/MusicTrigger.cs
FantasticGames1: The issue was that this wasn't working, and there were many layers of scripts it was connected to that were causing it not to work. I would have had to overhaul too many other components to make it work(which could cause more bugs). We needed something more straightforward and simpler, especially for such a simple task.

Assets/Resources/SaveData/Data.json
FantasticGames1: Oh, i missed that. thanks

RyanASwanson (changes requested):
I feel incredibly mixed about approving of this. I truly believe we should be updating the old system and creating entirely new functionality in it's place. I understand that makes the scope of the task larger, but I feel that should be done. We can adjust your scope to make sure it gets in. Creating an entirely new system creates more issues than it solves. I cannot approve of a system that still needs changing

PR #761, opened 2025-04-05
Gold des implement demo context lv 1630

C-Low01:
Demo context found in the demo context scene. The video is currently a black screen, but can be replaced with a still image if desired.
Text for demo context is currently formatted to fit the lower portion of the screen. If we want it in the middle it will need reformatting.


SymphonicSweet (approved):
The context moves a little quickly into the first cutscene, but other than that, it looks good. Update from gold and merge

PR #762, opened 2025-04-06
Gold eng fix or standardize game screen scaling lv 1741

nrice25:
I have probably missed things; ping me if need be, but I am also passing this off to Adam for cleanup

A non-code change to allow for proper UI scaling of different UI assets;
This does not affect the main menu
It does affect:
- Different on-screen health effects
   - The heart
   - The dmg veins
   - The green health effect
- The objective (it isn't properly placed but I'm exhausted)
- The pause menu

The harpoon has hard coded it's UI elements (My current recommendation is to add the ammo to the actual UI space instead of instantiating it through code, so as to better scale with the rest of the UI)


hhlaxbro (commented):
Sorry for the wait. I just checked it. there is a bit of scuffness with the pause menu especially, and selecting options from there. That said from what I was able to test everything else looks solid! 

The biggest thing that will be a challenge for you and @Adam Garwacki is making sure there is consistency in menu navigation. That means constant size, placement of repeating elements, and so on. That said functionality needs to be in first so keep doing that but yes things will need to be adjusted to the appropriate sizes and spots throughout the game

PR #764, opened 2025-04-06
Brightness slider is not interactable LV-1514

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-1514
changed the slider element to interactable and changed the colors to full alpha so it looks like the other sliders

also a bunch of silly fmod ghost changes grrr


RyanASwanson (approved):
Yup! Approved, and good to merge

PR #765, opened 2025-04-06
Fix title screen sliding based on screen size LV-1752

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-1752
turns out not only does the time value in smooth damp not need to be adjusted based on screen size it's also already frame independent so it doesn't need time delta time (unlike what the video I watched said) 
anyway it's fixed now


RyanASwanson (approved):
Tested it a good number of times and I see no issue. Though the current time being set to 30 feels excessive.

PR #766, opened 2025-04-06
Gold UI screen scaling setting design lv 1735

SolarFoxG:
Adjusted nearly all UI elements outside of the PlayerHUDCanvas (which involves in-game subtitles, heart, key item icons, damage vignette, the reticle, ammo icons, and the pause menu) to accommodate various screen sizes.

To test this: swap to various resolutions in the Editor, especially custom widescreen presets. Stretch the screen. See if things break.


hhlaxbro (changes requested):
Great work Adam! This seems super well worked! The only issue I saw was that the controls images in the controls menu was super small. Also the ESC in the options menu is notable smaller and placed lower on the screen than in the controls. If you could make the ESC the same/similar to the one controls menu that would be best. Don't forget the settings one to. Other than that it looks awesome to me! Great work!

hhlaxbro (approved):
Approved

PR #770, opened 2025-04-08
gold-eng-sprint-animation-functionality

RyanASwanson:
(Currently on the task list so there is no Jira ticket. Instead here's the task list: https://docs.google.com/spreadsheets/d/1LWLiJSVBJ8q5QARLaoSWM-zSfN-sJnJQUgWlbKiEPgg/edit?gid=288836653#gid=288836653)

Nothing should feel like it has changed through this PR. Basically just added functionality so that an animation can change on the harpoon gun when the chase sequence starts. If you would like to test it in engine you can open the animator on the harpoon gun, and you will see a boolean called 'sprint' be changed to true once the chase sequence starts

In Engine Review Estimation: <5 minutes (Getting to the chase sequence takes a bit)


nrice25 (approved):
Approved codewise; I don't think this is enough of a change to look at this in engine

PR #772, opened 2025-04-08
gold-eng-monster-spawn-vfx

RyanASwanson:
Jira task doesn't exist yet so here is the task list:
https://docs.google.com/spreadsheets/d/1LWLiJSVBJ8q5QARLaoSWM-zSfN-sJnJQUgWlbKiEPgg/edit?gid=288836653#gid=288836653

The goop vfx plays when a tentacle emerges from a wall. Double triple check that the orientation of the vfx is in the right direction regardless of where it spawns. I checked thoroughly but if anything were to go wrong in this PR it would be this.

In Engine Review Estimation: <6 minutes (To thoroughly check the orientation)


nrice25 (approved):
Goop spawns in; approved

astapay (approved):
Yippee!

PR #775, opened 2025-04-08
Gold des implement monster lv 1536

BreadReaver:
Adjusted whack prefab to fit into hallways
Placed new monsters into scene


TurquoiseDragon (approved):
Everything seems to work and is up to snuff until we get fixes for the animations

MagicTurtle013 (approved):

PR #777, opened 2025-04-08
Gold des implement book ripped to shreds lv 1728

SymphonicSweet:
Added bloodied metal pipe (near the third Vernon note), the ripped book (near the giant tipped boat), and the map asset (on top of a pallet in the chase sequence build up).


pmetcalf2 (approved):
Looks good, go ahead and merge

PR #779, opened 2025-04-09
gold-eng-infinite-ammo-restock-prefab

RyanASwanson:
Not on Jira. Here's the Task sheet : https://docs.google.com/spreadsheets/d/1LWLiJSVBJ8q5QARLaoSWM-zSfN-sJnJQUgWlbKiEPgg/edit?gid=288836653#gid=288836653

Made the ability to have an infinite ammo rack. This is for the firing range room. It will have a different model when design works with it. I made a scene called AmmoBoxTest to test it out. Yes the ammo box is just floating in the firing range, I do not care. Yes the harpoons are clipping in the model. I also do not care. That's a design thing. I just got it functional.


nrice25 (approved):
Works in engine, approved

astapay (approved):
Yippee!

PR #781, opened 2025-04-09
Gold des tutorial door lv 1822

MagicTurtle013:
I have added the door to the tutorial and it opens after putting down the book.

Engineers- 
I updated the TutorialPopUp.cs script to add put-down functionality because it didn't have it originally. I took about 4 lines of code from the NoteInteractable.cs script so I can add unity events to the tutorial interactable so I could get it to interact with the StoryManager. So it should be a quick review.

Designers-
All you need to do is to run through the game and check to see if I messed up anything lol. 
I had to mess with the story manager and all the scenes because the tutorial is the new "0" on the StoryManager so I had to basically add 1 to all the props that moves the manager forward. It was fine when I ran though, but doesn't hurt to triple check.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-1822?atlOrigin=eyJpIjoiNTA1N2EwOTNlMWMwNDI4NTliNGU2NGQzYTc0NDFmZDciLCJwIjoiaiJ9


nrice25 (approved):
Approved, code looks good

pmetcalf2 (approved):
Door is opening on tutorial put down, your still able to run through the game, go ahead and merge.

PR #782, opened 2025-04-09
Gold des create all dialogue scriptable objects lv 1820

pmetcalf2:
All Scriptable Objects for the dialogue have been made inside the below-decks dialogue folder. They are not timed due to not having a VA in the game yet. Check to ensure the lines match the correct number on the script and that nothing got cut off in putting all the dialogue in.


C-Low01 (approved):
I got a lot of work to do now

PR #783, opened 2025-04-09
gold-demo-note-functionality

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1869

There is a test scene called DemoTest. Design requested the demo scene (the one after pressing play that provides context) to be able to be progressed through. I asked if just using note functionality would do and they said yeah. Basically just used that with some minor tweaks, and added a button on the last page (with controller support) that progresses to the intro cutscene. Design/UX will make a pass through it at some point. Note that this scene is currently only for testing but might just get swapped in after the visuals pass.

In Engine Review Estimation: <6 minutes (If you also test controller support)


Assets/Scripts/UIScripts/InGameMenuSwap.cs
```diff
@@ -16,14 +16,12 @@ public class InGameMenuSwap : MonoBehaviour
...
-    private EventSystem _eventSystem;
-    
     /// <summary>
     /// Makes UI buttons selected when a new UI screen pops up from gameplay
     /// </summary>
     private void OnEnable()
     {
-        _eventSystem = FindObjectOfType<EventSystem>();
-        _eventSystem.SetSelectedGameObject(_firstSelectedButton);
+        // EventSystem.current uses the event system that is in use
```
nrice25, line 24: I am glad you made this change :)

nrice25 (approved):
Approved, works on keyboard and mouse; I did not check controller. However, because the code is relatively the same as before, I'm not worried.

astapay (approved):
Yippee!

PR #788, opened 2025-04-10
Changed Harpoon Textures to look not glossy

fiendlyghost:
Imported new texture maps and made a new material for the harpoon and just attached it to the harpoon. (within the harpoon prefab)


TurquoiseDragon (approved):
Game works and the harpoon look good!

PR #791, opened 2025-04-10
gold-eng-Note-rustling-LV-1864

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1864

**REQUIRES CONTROLLER**
Before this PR if you interacted with a note (on controller), and then pressed the interact again it would close and instantly reopen the note all in 1 frame, and would play the note opening audio again. The task was originally just to fix the repeating audio but I uncovered the root cause in this and fixed that instead. Reason why this happened is that open note and close note are both bound to the same button so the player would close the note, interact, and reopen the note.

How to test: Go to any note while using a controller and keep spamming X or whatever the button is for a non playstation controller. (The bottom one). Now the note will close as it is intended to do.

In Engine Review Estimation: <3 minutes


nrice25 (approved):
Approved, the fix works in engine

PR #792, opened 2025-04-10
Gold-eng-adjust-default-settings-lv-1882

Lazydev114:
Jira: https://bradleycapstone.atlassian.net/browse/LV-1882
Description: Adjusted default setting to wanted values. Test by playing without any saved data and checking settings values


Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/SaveManager.cs
```diff
@@ -72,6 +72,14 @@ private void GameplayStartingValues()
 
         // This sets the initial scene to 1 because it is the game scene (the title scene is 0)
         _gameSaveData.SetCurrentSceneIndex(1);
+
+        //sensitiviy starting values
+
+        // Convert the sensitivity to a string
+        string _settings = 225 + " " + false + " " + false;
```
RyanASwanson, line 79: Perhaps this could be calculated by using the max value. That way if we change the max value it still gets the midpoint to start with?

Assets/Scripts/UI/GameplaySettings.cs
```diff
@@ -24,7 +24,15 @@ public class GameplaySettings : MonoBehaviour
 
     [SerializeField] private string _GameplaySettingFilePath;
 
+    public static GameplaySettings instance;
 
+    private void Awake()
```
RyanASwanson, line 29: XML comment

Assets/Scripts/UI/GameplaySettings.cs
```diff
@@ -24,7 +24,15 @@ public class GameplaySettings : MonoBehaviour
 
     [SerializeField] private string _GameplaySettingFilePath;
 
+    public static GameplaySettings instance;
```
RyanASwanson, line 27: This is public so should start capitalized

Assets/Scripts/UI/GameplaySettings.cs
```diff
@@ -24,7 +24,15 @@ public class GameplaySettings : MonoBehaviour
...
 
+    private void Awake()
+    {
+        if (GameplaySettings.instance == null)
```
RyanASwanson, line 31: A trick that Nick found was that instead using the function .IsUnityNull is actually more efficient that checking ' == null'

Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/SaveManager.cs
```diff
@@ -199,7 +207,9 @@ public override void SetUpMainManager()
     {
         base.SetUpMainManager();
         EstablishPath();
+        StartingValues();
```
RyanASwanson, line 210: This should not be here. The load function directly under this one will call the starting values if there is no save data so this is already done. Even if that wasn't the case the changes made from this function are instantly overwritten from the Load function so it doesn't change things.

RyanASwanson (changes requested):
The brightness settings slider appears to be at 25% by default. The sensitivity slider seems to be around ~60% by default. That seems to be an issue on the side of the slider since the settings do seem to be reset properly. Also I left some general cleanup comments.

astapay (approved):
Hooray!

PR #793, opened 2025-04-10
lv-gold-monster-encounter-music-not-playing-LV-1743

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1743

Adjusted the music trigger to be more versatile and adjusted the enemy prefab and enemy whack-a-mole prefab to accomodate for those changes. For some reason the EnemyEncounterMusic in particular doesn't work. **When testing this just swap out the enemy music for the TitleMusic.** Then just get to a point in which you are fighting an enemy and stand within the range. It will switch music tracks. Music switches a bit slow but that's up to design to tweak to what they like. (You should also hear the ambient bug noises as currently that is the temp audio asset used for the breathing level associate with combat)


Assets/Scripts/UniversalFunctions/MusicSwapPlayerTrigger.cs
```diff
@@ -54,36 +66,51 @@ private void Start()
...
     {
-        PersistentAudioManager.Instance.StartMusicByID(_musicID);
+        PersistentAudioManager.Instance.StartMusicByID(_musicEnterID);
     }
 
     /// <summary>
     /// Switches the volume of the music
     /// </summary>
-    private void SwapVolume()
+    private void SwapVolume(float volume)
```
nrice25, line 111: Pardon my french, but you aren't using the float that you added to this?

Assets/Scripts/UniversalFunctions/MusicSwapPlayerTrigger.cs
```diff
@@ -54,36 +66,51 @@ private void Start()
...
+            //Using an Else If just in case we end up adding more EMusicTriggerTypes
+            SwapVolume(volume);
         }
     }
 
     /// <summary>
     /// Switches the music to play
     /// </summary>
-    private void SwapMusic()
+    private void SwapMusic(int id)
```
nrice25, line 103: Why are you adding in a parameter for the music id if you aren't using it?

Assets/Scripts/UniversalFunctions/MusicSwapPlayerTrigger.cs
RyanASwanson: It has 2 floats for this up above. One for entering collision and one for exiting. We are calling the function using one of the 2 floats as inputs.

Assets/Scripts/UniversalFunctions/MusicSwapPlayerTrigger.cs
RyanASwanson: Same logic as my previous reply. There is an int for entering collision and one for exiting, then we are calling this function with either of those 2 as inputs

nrice25 (approved):
It works in engine! Approved!

PR #795, opened 2025-04-11
gold-eng-Player-can-backtrack-after-continuing-LV-1886

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1886

Save points can now have objects get enabled when you load from that point. I made a scene called SavePointTest as an example. If you walk up really close to the key there is a save point. You can then restart the scene and you will load at that point, with an extra surprise waiting behind you.


astapay (approved):
Yippee!

PR #799, opened 2025-04-11
Separate pause menu from player hud canvas lv 1889

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-1889

created a new canvas and prefab for the pause menu separate from the player hud so they can be adjusted for screen size separately. I expected this to break things but from my testing everything seems to work fine fingers crossed.


hhlaxbro (approved):
Looks good to me. That said, wait till Adam is able to get his eye on this to approve as he is the one who needs this new version, and needs to make sure that it will suit his purposes

PR #801, opened 2025-04-11
Sound Tube Prefab is in and functional

C-Low01:
Totally my first try

Theres a folder called CleosSoundTubes which has the updated prefab in it, as well as one example of it working in the key room. I will be adding the rest into the game during the Sound Tube VA task


TurquoiseDragon (approved):
It works but audio overlaps

PR #802, opened 2025-04-11
gold-art-pipe-assets-lv-1762

MilesRogers3D:
Created a modular kit of pipes and placed an example in the first maze

image

- Assets are located under Assets/Art/3D/Environment/PipeKit


BreadReaver (approved):
Miles's pipe looks good

PR #804, opened 2025-04-12
Gold eng screen scaling setting functionality lv 1736

nrice25:
[Jira Ticket](https://bradleycapstone.atlassian.net/browse/LV-1736?atlOrigin=eyJpIjoiMjUzYmM1MTQzZWY3NDczNGJmNTM3MzM4ODU4YjU1NzkiLCJwIjoiaiJ9)

**Test this by making a build and going to the video section under settings**

This task may take around 15-30 minutes for testing (UI is kinda, not great ~ at least on the main menu -> when changing resolutions)

There is currently a problem with there being 3x each resolution; but I wanted to get this in a PR first, if you have any ideas on how to fix this it would be appreciate (This problem doesn't exist in editor)


Assets/Scripts/UI/VideoSettingsBehaviour.cs
```diff
@@ -78,4 +100,101 @@ public void ToggleGoreSetting()
...
+        Screen.fullScreen = !Screen.fullScreen;
+    }
+
+    
+    #region Resolution Functions
+
+    /// <summary>
+    /// This changes the player's resolution
+    /// </summary>
+    private void ChangeResolution(int newResolutionPointer)
```
RyanASwanson, line 118: Don't forget to include the integer in the XML comment

Assets/Scripts/UI/VideoSettingsBehaviour.cs
```diff
@@ -78,4 +100,101 @@ public void ToggleGoreSetting()
...
+            // Grabs the name of the resolution
+            tempOption = resolution.ToString();
+            // Cuts out the refresh rate
+            tempOption = tempOption[..(tempOption.LastIndexOf('@') - 1)];
+            // And then adds it as an option to a list
+            TMP_Dropdown.OptionData testerOptionData = new TMP_Dropdown.OptionData(tempOption);
+            if (!optionData.Contains(testerOptionData))
+            {
+                Debug.Log(testerOptionData.text + " " + testerOptionData);
+                Debug.Log(optionData.Contains(testerOptionData));
```
RyanASwanson, line 150: We could remove these debug statements

Assets/Scripts/UI/VideoSettingsBehaviour.cs
```diff
@@ -78,4 +100,101 @@ public void ToggleGoreSetting()
...
+    private void StoreWidthHeight(string resolution)
+    {
+        int seperatorIndex = resolution.LastIndexOf('x');
+        _resolutionWidth = int.Parse(resolution[..(seperatorIndex-1)]);
+        _resolutionHeight = int.Parse(resolution[(seperatorIndex+2)..]);
+    }
+    
+    #endregion
+
+    private void OnDisable()
```
RyanASwanson, line 196: XML. Built in functions we are generally more relaxed on not having XML comments but still should do it regardless

Assets/Scripts/UI/VideoSettingsBehaviour.cs
```diff
@@ -65,7 +87,7 @@ public void ToggleSubtitleSetting()
     {
         SaveManager.Instance.GetGameSaveData().IsSubtitlesOn = _subtitleToggleButton.isOn;
         //stop any current subtitles
-        if (FindObjectOfType<DialoguePopUps>() != null)
+        if (!FindObjectOfType<DialoguePopUps>().IsUnityNull())
```
RyanASwanson, line 90: You didn't make this one, but we really should switch this off FindObjectOfType. I can't really request that be done in this PR, as that would just push things back, but if I delay it, it probably wouldn't be done. I suppose keep it in mind, but feel free to carry on.

RyanASwanson (approved):
Code is pretty much fine, with some minor adjustments requested. Obviously there are some oddities regarding things scaling weird, but that is Adam's responsibility so I'll look past that for now. Settings do seem to work properly. I know there is the duplicated resolutions you are looking at now. I'll double check the PR after that, but otherwise it's fine. I know production is getting increasingly picky about not letting things through that feel incomplete, but I think this is the right call.

Assets/Scripts/UI/VideoSettingsBehaviour.cs
nrice25: I'm not sure if this can be easily handled outside of FindObjectOfType?
Because this exists in both the main menu and the game scene, so (to me) there isn't a one size fits all

RyanASwanson: My thought process would be an event that this could just invoke, and then if we are in a scene with DialoguePopUps it would just subscribe to that. Regardless this is nitpicky, and not mandatory

RyanASwanson (approved):
Pretty much my opinion hasn't changed. Scaling is still weird, but that remains Adam's responsibility. The issue of duplicate resolutions is fixed though. So still approved

MagicTurtle013 (approved):

RyanASwanson (approved):
Looks good, scales well

PR #805, opened 2025-04-12
Gold des implement new va recordings lv 1852

C-Low01:
All new va that utilizes the story system, and the note pickups are in the game.
VA that is not in this branch: 
Triggered va
starting cutscene va
ending cutscene va

Known issues: 
This line is too long, and exceeds the max subtitle viewing time of 8 seconds. It needs to be cut into 2 different lines to display properly: “Finally finished down there? I can hear the generator from up here again. Go ahead and come on up. I’m doing a quick check up here before we shove off again in case anything else broke.”

This line was cut too early, and has the actor reading the line number before saying the line: “About time.”


TurquoiseDragon (approved):
Lines are in but need some adjusting for polish

PR #806, opened 2025-04-13
Gold art rain patch vfx lv 1764

noodlecat519:
VFXRainPatch prefab is in Art/VFX/WorldEffects. You can see it in the VFXWork scene


angelkait (approved):
you're on a roll caelie! this was awesome, great work thank you!

PR #807, opened 2025-04-13
Added triggered dialogue VA

C-Low01:
Refer to title, mostly.
What ISNT in yet:
A few banter lines, between the player and captain. This is because there's no way to space the triggers out in a way that would make the lines flow naturally. This can be solved by either: Making banter lines a triggered event directly in the story manager, or having both lines (Captain dialogue + Player response) recorded in the same sound clip, so that they can play in sequence without the need for 2 triggers.


TurquoiseDragon (approved):
Needs iteration but it works and we should get more conversation on this soon.

PR #809, opened 2025-04-13
gold-des-Implement-Light-Bugs-VFX-LV-1837

ogNINJAGOLDBOX:
Added Light Bugs to all the scenes. See if they are noticable or if any of them are used in ways that are not desired

https://bradleycapstone.atlassian.net/browse/LV-1837

Basically just play and see if you notice them. After that you can look in the scene view and see if there are ones you would never notice or when playing if there are any that look off.

I also wasn't sure if we cared about consistency between scene changes, usually you are not looking at the light when it changes and I thought it could kind of hint at the change, but if we'd rather the room you are in to stay fully consistent I can change that


Dusk553 (approved):

MagicTurtle013 (approved):

noodlecat519 (approved):
Awesome job with the bugs :)

PR #815, opened 2025-04-13
Trigger main menu animations lv 1862

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-1862

added some variables and a tiny bit of code to trigger Hayden's title screen animations at specific (customizable) times during the title screen scroll.


hhlaxbro (approved):
It looks siiiick thank you Jeremiah 🔥🔥🔥

Assets/Scripts/MenuFunctionality/TitleScreenScrolling.cs
```diff
@@ -60,6 +68,16 @@ private IEnumerator ScrollingScreen()
                     _screenScrollTime);
                 yield return null;
 
+                if (transform.position.y / _movingDestination.transform.position.y >= _skullTriggerPercent/100)
+                {
+                    _skullAnimator.SetTrigger("StartMoving");
```
nrice25, line 73: My only critique is that you could make the trigger string a constant variable

nrice25 (approved):
Approved, code looks good and I heard you getting glazed at the meeting over this; so I'm not worried about in engine

Assets/Scripts/MenuFunctionality/TitleScreenScrolling.cs
```diff
@@ -60,6 +68,16 @@ private IEnumerator ScrollingScreen()
...
+                }
+
+                if (transform.position.y / _movingDestination.transform.position.y >= _menuTriggerPercent / 100)
```
RyanASwanson, line 76: We could probably save the calculation of the left side of the if statement into a float as we are doing that same calculation in both if statements. Also we could probably do the calculation of dividing the 2 percent variables by 100 in start, it doesn't really need to happen every frame of the main menu scroll as it's never going to change

Assets/Scripts/MenuFunctionality/TitleScreenScrolling.cs
```diff
@@ -60,6 +68,16 @@ private IEnumerator ScrollingScreen()
...
+                {
+                    _menuAnimator.SetTrigger("StartMoving");
```
RyanASwanson, line 78: Tried running this and I put in print statements just to double check it was working properly. Only thing I encountered is that these SetTriggers are going to happen every frame until the coroutine ends after they meet their condition. So long as Hayden is careful not to have the Animator set up to transition from AnyState into the animation then it should be good, but he should be made aware of that.

RyanASwanson (approved):
Overall functionality is good, left some notes about clean up that would be nice if they were either resolved or responded to.

PR #817, opened 2025-04-13
Gold des submenu duplication fix lv 1823

SolarFoxG:
Jira: https://bradleycapstone.atlassian.net/browse/LV-1823

Documentation: N/A

How to Test: In the editor, try to make directional inputs and select menu option while a submenu (notes, settings, or controls) is open. If this PR has been done right, the player should be unable to explore nested menus while a submenu is open, meaning nothing can accidentally be selected without the player knowing. The credits scene has also been made fully non-additive as part of fixing this bug.


RyanASwanson (approved):
Yup, all good

PR #818, opened 2025-04-13
Gold eng adjust default settings 3 lv 1882

Lazydev114:
Jira: https://bradleycapstone.atlassian.net/browse/LV-1882
Description: Adjusted default setting to wanted values. Test by playing without any saved data and checking settings values


RyanASwanson (approved):
All good. Only thing to note is the brightness slider doesn't seem to be at 50% by default. Maybe a similar thing that was done with the sensitivity slider needs to be done with the brightness slider as well. Currently our max brightness is 2, so a default brightness of .5 isn't actually 50%.

PR #824, opened 2025-04-14
gold-des-Make-and-Implement-Running-Gun-Animation-LV-1801

ogNINJAGOLDBOX:
Added a sprint animation and implemented it

https://bradleycapstone.atlassian.net/browse/LV-1801

Just need to play the chase sequence and it should work just fine. Let me know how you feel about the animation


astapay (changes requested):
Animations are disjoint when you try to do something else while sprinting

MagicTurtle013 (approved):

TurquoiseDragon (approved):
It works with a low amount of jank

astapay (approved):
Yippee!

PR #827, opened 2025-04-14
Gold art create navigation decals lv 1768

noodlecat519:
All navigation decals can be seen on the wall in the MazeDecalTesting scene


hhlaxbro (approved):
Looks cool!

PR #831, opened 2025-04-14
gold-eng-Lighting-Related-Lag-LV-1890

astapay:
So this task ended up not really being fixing any lighting. We determined that the looking issue was not with the lights but rather with the camera rendering game objects that weren't even in view. So, I just turned down the far clip plane for the camera. I still took this opportunity to clean up some of the Light Controller code though. Just make sure the Light Shifting still works ig?

Code Review Time: <1 minute
In-Engine Review Time: <5 minutes


RyanASwanson (approved):
Yippee

PR #835, opened 2025-04-15
Lv 1964 oliver hotfix patch

BreadReaver:
Deleted final monster location in hallway to spanner room

Removed collider on upside down tables right before big crate room so you dont clip head on them

Put a metal separator over any corner that had a hole

Changed the boat man to be greener since him being orange felt weird

Added dummy holes throughout maze

Added SFX to Datamoshing effect

Added SFX to jumpscare 

Deleted that red lifter object in the first hallway of the starting area

Took datamoshing out of the chase sequence

Moved the walls down in the swamp anomaly to try and prevent you from seeing the void


Primalwaif5691:
It looks good and sounds good on my end

Primalwaif5691 (approved):

PR #836, opened 2025-04-15
gold-eng-Harpoon-Racks-Dont-Work-LV-1971

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1971

Updated the logic for the harpoon racks knowing if the ammo is at max (this took me longer than it should've I am stupid), made it so that loading into a new maze segment while not at max ammo has the harpoon racks be interactable.

Try shooting a few times, then progress into maze 2 (Don't shoot at this time as that is what fixed the bug in the past) until you reach the first harpoon rack. You can interact with it properly (The first rack is right past the swamp anomaly.


nrice25 (approved):
Approved, it works; just make sure to update from main

PR #837, opened 2025-04-15
gold-des-Update-Med-Kit-and-Harpoon-Assets-LV-1813

ogNINJAGOLDBOX:
Updated the assets for the Ammo Rack and Healing Kit so that they have stuff that they hang on rather than floating

https://bradleycapstone.atlassian.net/browse/LV-1813

Just play and look at them


Dusk553 (approved):

TurquoiseDragon (approved):
Looks good!

PR #840, opened 2025-04-15
UI click sfx functionality lv 1888

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-1888

adds a click sound when you press buttons in the UI

Known issues:
sounds do not play in pause menu (but it is set up to work once the audio pausing issue is fixed)

also for some reason pressing escape in the settings and controller screen plays the clip twice (I have no idea why escape is playing the sound bc it doesn't do that in the quit confirm screen)


RyanASwanson (approved):
Functionality works properly, just left some notes about comments, and naming conventions. For the most part I disagree with some of Adam's notes, as they cover things closer to implementation and audio, though I do agree about the note of using directional keys to also play audio

hhlaxbro (approved):

astapay (approved):
Yippee!

Assets/Scripts/MenuFunctionality/TutorialPopUp.cs
```diff
@@ -35,6 +35,8 @@ public class TutorialPopUp : MonoBehaviour, IPlayerInteractable
     private bool _hasDoorOpened;
     private InGameMenuSwap _menuSwapScript;
 
+    [SerializeField] private ButtonSFXManager ButtonSFXManagerReference;
```
RyanASwanson, line 38: This is private so should follow the private naming convention

Assets/Scripts/MenuFunctionality/TutorialPopUp.cs
```diff
@@ -93,6 +95,8 @@ public void OpenTutorialPopUp()
     /// <param name="pageChangeAmount">The value to change the page by</param>
```
RyanASwanson, line 95: This isn't covered in the changes so I cannot place the comment directly on it. But for the line 
TimeManager.Instance.PauseGameToggle(true);
Can we just set that to false real quick. Makes the ui sounds not work if it's true

Assets/Scripts/MenuFunctionality/TutorialPopUp.cs
```diff
@@ -93,6 +95,8 @@ public void OpenTutorialPopUp()
...
     public void ChangePage(int pageChangeAmount)
     {
+        ButtonSFXManagerReference.PlayClickSFX();
```
RyanASwanson, line 98: It seems to play the audio even if you cannot move in that direction. This applies to the note too

RyanASwanson (approved):
Just some notes on bug fixes and clean up

Assets/Scripts/UI/ButtonSFXManager.cs
```diff
@@ -0,0 +1,20 @@
...
+// Author:          Jeremiah Peters
+// Creation Date:   4/14/2025
+//
+// Description:     Enables menu buttons to play sound effects
+******************************************************************************/
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class ButtonSFXManager : MonoBehaviour
```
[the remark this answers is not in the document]
JeremiahPeters311: added the class header

i guess calling it a manager is hyping it up a bit but it does indeed manage button sfx. i dunno seems reasonable to me

RyanASwanson (approved):
Yup! Approved!

PR #846, opened 2025-04-15
Second-check-on-dialogue-LV-1977

pmetcalf2:
Fixed line 64 player to no longer have the old placeholder dialogue(i lied, hahahaha), and fix the error that was stated on Line40 player for its name. Also checked to make sure no other objects had this problem so we should be good.


TurquoiseDragon (approved):
Everything that was intended in the PR is in

PR #857, opened 2025-04-17
bug-window-asset-should-have-a-dark-skybox-like-the-title-screen-LV-1980

MilesRogers3D:
Made window background darker to fit with the intro cutscene, fixed "black box" issue with the camera bounds

A few minor code changes:
- Got rid of the global registry/distance check in FakeWindowManager.cs since it just overcomplicated things
- Added an extra term in the view angle calculation to account for which direction the window is facing in world space (aka fixing the black box issue)
- Edited FakeWindowObject.cs to work with the changes in FakeWindowManager.cs


Primalwaif5691 (approved):

RyanASwanson (approved):
Yeah code is simplified down, and there are no issues in engine.

PR #858, opened 2025-04-17
Gold eng implement monster animations lv 1880

troberts2:
Everything is testable for the new whack a mole monster in the ProceduralTest scene. Should look way better now and I have yet to notice any bugs with it. Shouldn't take more than 20 minutes to run through the code and test it.


RyanASwanson (changes requested):
Not sure why but the monsters get very jittery as they appear from the ground. That is a major problem and needs to be fixed before this is merged

Assets/Art/3D/Boss/BossScripts/Attacks/ProceduralAttack/ProceduralVine.cs
```diff
@@ -85,6 +86,10 @@ public enum EVineState
     private void Start()
```
nrice25, line 86: Having an xml comment here would be nice

Assets/Art/3D/Boss/BossScripts/Attacks/ProceduralAttack/ProceduralVine.cs
```diff
@@ -324,17 +329,49 @@ private IEnumerator WSnapAttack()
     {
         yield return new WaitForSeconds(_waitAfterRearBackTime);
 
+        //setup and trigger attack
+        _animator.SetTrigger("bite");
```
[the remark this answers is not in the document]
nrice25: Overall, having the non-named number values in this section be class variables would be nice;
also for ease of changing them

Assets/Art/3D/Boss/BossScripts/WeakPoints/WeakPointHandler.cs
```diff
@@ -194,7 +190,7 @@ private void MaxWeakPointsDestroyed()
         {
             if (_proceduralVine != null)
```
nrice25, line 191: ! .isUnityNull

Assets/Art/3D/Boss/BossScripts/Attacks/ProceduralAttack/ProceduralVine.cs
```diff
@@ -128,7 +140,7 @@ private void Update()
         //whack a mole vine
         else
         {
-            //appearing
+            /*//appearing
```
nrice25, line 143: Because this won't be in; it's safe and recommended to delete this as it can be found within a past PR

RyanASwanson (approved):
Code is improved. I do worry about how the monster works when coming out of the ground/ceiling. We have to consult Oliver for that one, so hold off merging

PR #861, opened 2025-04-17
gold-eng-Harpoon-freezes-in-air-LV-1987

RyanASwanson:
https://bradleycapstone.atlassian.net/browse/LV-1987

If the player died and resumed the game or left to go the the main menu and returned to gameplay the harpoons would get stuck in the air and not move.
That should be resolved here. The harpoons were confused by what the main camera was as the fake window had the tag and it never updated the variable storing the main camera.

Code Review Estimation: <1 minute
In Engine Review Estimation: <6 minutes (Be thorough)


TurquoiseDragon (approved):
It no longer freezes!

nrice25 (approved):

PR #863, opened 2025-04-18
Update scriptable objects LV 1988

pmetcalf2:
All scriptable objects have there get time before display set to 0.


SymphonicSweet (approved):
checked them all, they all are set to 0. Update with gold and merge

PR #867, opened 2025-04-19
Gold des menu hover state update lv 1847

SolarFoxG:
Jira Task: https://bradleycapstone.atlassian.net/browse/LV-1847

Expected Test Time: 10 Minutes

How to Test: When accessing the settings menu either from the main menu or the pause screen, ensure that you can toggle all options using mouse, keyboard, and controller inputs. The menu should be fully navigable using only button prompts. The current option you are locked onto or hovering over should always be bright, and other menu assets should be darkened. Please ensure that settings values, particularly toggles, are accurate to their effects in-game.


JeremiahPeters311 (approved):
👍

PR #868, opened 2025-04-19
generator sounds in story manager lv 1828

pmetcalf2:
added the generator breaking and the generator successfully starting inside the story manager. The second sound is much quieter, its the sound itself.


C-Low01 (approved):
Works!

MagicTurtle013 (approved):

PR #876, opened 2025-04-20
gold-eng-Gun-doesn't-reload-if-empty-while-picking-up-new-ammo-LV-1994

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-1994

Reload works on auto reload when picking up ammo while empty


nrice25 (approved):
Approved, works in engine

PR #877, opened 2025-04-20
gold-art-fix-water-shader-lv-1763

MilesRogers3D:
Made some edits to the water material used in maze 2 to fix any visual glitches and make it look more like water than a translucent plane


BreadReaver (approved):
Sure

PR #883, opened 2025-04-20
Gold des implement chase sequence running sfx lv 1839

Dusk553:
Added in the funtionality for the chase audio, which should loop when Scott Updates the Fmod audio.
Jura Task:https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=63c32962c52a63dcda83f747&selectedIssue=LV-1839

You can check in the persistent audio manager and the storymanager in the 7th set of story parts


SymphonicSweet (approved):
Looks good, if you can increase the volume and get it to loop at some point, do so, but other than that I like it

BreadReaver (approved):
scary!

PR #890, opened 2025-04-21
gold-art-fix-lights-vanishing-lv-1975

MilesRogers3D:
Adjusted the light ranges so that Unity doesn't get confused about which point lights we can see/run up against its internal limit. This also improved the performance a little bit


astapay (approved):
Yippee!

PR #896, opened 2025-04-21
Gold des finalize objectives lv 1826

SymphonicSweet:
start the chase sequence and see if the "run" objective comes up


pmetcalf2 (approved):
run pops up after the monster screams, looks good to me.

PR #897, opened 2025-04-23
Gold-qa-eng-story-manager-reset-LV-2023

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-2023

Issues kept compounding in this PR so a bit was fixed here. First it was the initial issue of the first story beat playing regardless of what save point you are playing from. Pretty simple fix, just have the game keep track of if you are playing on new save data, then when the game checks for if any of the story beats are supposed to play on awake you just have it not do so if you already have gameplay save data. Next there was an issue in which the first story beat would only sometimes play the dialogue. Reason why was an order of events thing, to fix this I made the DialogueSfxManager actually a manager, and therefore is set up before anything else (Also cleaner code, yippee). Then I found that new save data actually did not clear your current story beat that was saved. Minor oversight, just cleared it on new gameplay data. Then I found pressing new game, letting the intro dialogue play, and then going to main menu and pressing new game again the intro dialogue would not play.

How to test: Play the game on new save data, you will hear the first dialogue play every single time. Play through the game to the chase sequence (Apparently some of the save points got removed at some point on accident, but this one is still here). Die in the chase sequence and continue, and also reset the game and continue. In both cases you will no longer hear the intro dialogue, and will not see the pop up objective for reading the manual. 

In Engine Review Estimation: <12 minutes (if you are thorough and check all possibilities)


astapay (approved):
Yippee!

PR #898, opened 2025-04-23
gold-eng-Dialogue-Timing-Update-LV-2073

nrice25:
[Jira Link](https://bradleycapstone.atlassian.net/browse/LV-2073?atlOrigin=eyJpIjoiMDYxYmQ0NTdmZTg3NDNkNTlkZjI3NmM0M2ExOTZmYWUiLCJwIjoiaiJ9)

Go into engine and change the dialogue info "time before text" to different numbers depending on preference

This one is for you Peyton

It should take about 5-10 minutes to check in engine
<1 minute for code review

### DIFF
--- a/Assets/Scripts/ScreenText/DialoguePopUps.cs
+++ b/Assets/Scripts/ScreenText/DialoguePopUps.cs
@@ -93,8 +93,7 @@ public class DialoguePopUps : MonoBehaviour
         foreach (TextAndTimerData dialogueInfo in moreDialogue.GetTextAndTimer())
         {
             UpdateSubtitleSettingState();
-            // Wait to start displaying the next text
-            yield return new WaitForSeconds(dialogueInfo.GetTimeBeforeText);
+            
             // Takes the display text and makes it invisible
             _textContainer.text = dialogueInfo.GetText;
             _textContainer.maxVisibleCharacters = 0;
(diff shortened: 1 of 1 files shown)


Assets/Scripts/ScreenText/DialoguePopUps.cs
```diff
@@ -141,6 +140,8 @@ private IEnumerator DisplayText(ScriptableDialogueUi moreDialogue)
                 _textBackgroundContainer.maxVisibleCharacters++;
                 yield return new WaitForSeconds(1f / typeSpeed);
             }
+            // Wait to start displaying the next text
+            yield return new WaitForSeconds(dialogueInfo.GetTimeBeforeText);
```
RyanASwanson, line 144: I see this was moved from being at the start before hand, to being afterwards. Perhaps the variable should be renamed since it is no longer for a delay before the text and instead is a delay after the text for the next one. Better yet, why not have 2? One for before and one for after? (Note that if you do both be sure to update the FindTotalTime function accordingly)

RyanASwanson (approved):
Looks good, works how it should. Left a single note about variable naming, but that's it.

pmetcalf2 (approved):
Design can easily change how long text displays on screen after being fully written out.

Assets/Scripts/ScreenText/DialoguePopUps.cs
nrice25: I just renamed it, I could add in the extra functionality; but I want to stay as close to the line of, not coding anything extra as possible

PR #899, opened 2025-04-23
gold-eng-Medkit-doesn't-work-LV-2024

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-2024

Med kit works properly now. Problem was that the med kit only updated if it could be used when the player took damage. If you have a medkit in a different maze segment it would be disabled when you got hit, then when you enter that segment since the medkit starts unable to be used it wouldn't update. Now when a medkit is enabled it checks what it's interaction status should be.

How to test: Easiest way is to get to maze 3 and use the medkit on the wall. (You gotta be damaged at least once first)


astapay (approved):
Yippee!

PR #900, opened 2025-04-23
gold-eng-Reimplement-Cutscene-Harpoon-Gun-Issues-LV-2000

astapay:
yay! I basically just copied and pasted everything from the last time I did this task. Hopefully it won't get undone this time.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2000?atlOrigin=eyJpIjoiMDI3MzlkYWNjZjJlNDBkNzk0MGRjNzhkNjZlMDIxYTAiLCJwIjoiaiJ9

Code Review Time: <5 minutes
In-Engine Review Time: <10  minutes


Assets/Scripts/PlayerScripts/PlayerMovementController.cs
```diff
@@ -369,7 +369,7 @@ private Vector3 HandleVerticalMovement()
     private void DetermineInputState()
     {
         //Check for if the input was started
-        if(_movementInput.WasPressedThisFrame())
+        if(_movementInput.ReadValue<Vector2>() != Vector2.zero)
         {
             DirectionalInputStarted(_movementInput);
```
RyanASwanson, line 374: Doesn't this mean that every frame the player is holding a movement value this function is called? So it thinks you are starting moving every frame? Yeah so I tested it an the footstep audio is wonky and the movement sway is now wonky.

RyanASwanson (changes requested):
The line change here is causing unintended effects

RyanASwanson (approved):
Approved, just fix something first. If you fully ADS then enter a cutscene the FOV stays at that zoom, it then gets fixed the next time you ADS. I had an error happen 1 time relating to the reticle and not being able to do something while disabled, but I was unable to replicate it

RyanASwanson (approved):
Yeah this looks good to me. Harpoon is no longer visible in cutscenes and camera zoom from aiming is not a problem either. Tested it a few times with both cutscenes.

PR #913, opened 2025-04-28
gold-eng-Game-Crashes-after-Switching-Mazes-LV-2111

astapay:
Wow! I just removed the part of the code that was causing the memory leak. It wasn't really that important in hindsight and hurt more than helped. The movement sway works fine as is imo.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2111?atlOrigin=eyJpIjoiNTI4YjA3Y2ZkODQ4NGQxMDliMTRmMTVmMDQ0MWZhODUiLCJwIjoiaiJ9

Code Review Time: <1 minute
In-Engine Review Time: <5 minutes (just make sure the stack overflow doesn't happen)


RyanASwanson (approved):
Yup I played through the game start to finish both in editor and build and I was not able to encounter any error (associated with this at least, errors are very much still there) or freezes

PR #919, opened 2025-04-28
Gold des adjustments to dialogue in story manager lv 2088

pmetcalf2:
Adjusted the dialogue timing in the story manager to flow better. Also changed the timing on line 21 and took out the quotation marks on line 64.


C-Low01 (commented):
Works  
image

PR #920, opened 2025-04-28
gold-des-Re-Implement-Weird-Save-Points-LV-2109

JamesBondMeow:
- check betamaze and betachasebuildup prefabs


BreadReaver (approved):
save points in

MagicTurtle013 (approved):

PR #925, opened 2025-04-28
Objective changing LV-2069

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-2069

fixes an issue where triggering a new objective pop in while one is already in progress would display the new text for a brief moment before sliding in properly.

I tried a lot of more elegant solutions for this but none of them worked seemingly because the unity animator is just slow and bad (even with exit time checked off) so in the end I had to do a bit of a janky fix where I turn off the visual components of the objective hud for a short period while waiting for the animator to catch up. Far from ideal, but it's the only thing I could get to work.

### DIFF
--- a/Assets/Scripts/UI/ObjectiveHudBehaviour.cs
+++ b/Assets/Scripts/UI/ObjectiveHudBehaviour.cs
@@ -23,7 +23,8 @@ public class ObjectiveHudBehaviour : MonoBehaviour
     [SerializeField] private Animator _objectiveHudAnimator;
     [SerializeField] private TextMeshProUGUI _objectiveHudText;
     [SerializeField] private TextMeshProUGUI _objectivePauseText;
-    
+    [SerializeField] private Image[] _objectiveSprites;
+
     // Cached variables
     private WaitForSeconds _animationWait;
 
@@ -33,6 +34,7 @@ public class ObjectiveHudBehaviour : MonoBehaviour
     private void Awake()
     {
         _animationWait = new WaitForSeconds(_objectiveLingerTime);
+        _objectiveSprites = GetComponentsInChildren<Image>();
     }
 
     /// <summary>
@@ -41,11 +43,27 @@ public class ObjectiveHudBehaviour : MonoBehaviour
     /// <param name="objectiveHudTextString">the text that goes on the ui</param>
     public void ActivateObjectiveHud(string objectiveHudTextString) 
     {
+        //reset to default state if the animation is already going
+        if (_objectiveHudAnimator.GetCurrentAnimatorStateInfo(0).IsName("ObjectiveSlideIn"))
+        {
+            _objectiveHudAnimator.SetTrigger("AbortSlide");
+            _objectiveHudAnimator.ResetTrigger("SlideOut");
+            StopCoroutine(WaitForAnimation());
+        }
+
+        foreach(Image h in _objectiveSprites)
+        {
+            h.enabled = false;
+        }
+        _objectiveHudText.enabled = false;
+
+        //update text
         _objectiveHudText.text = objectiveHudTextString;
         // Mirrors info in pause menu
         SetPauseMenuObjective(objectiveHudTextString);
-
+        //starts slide animation
         _objectiveHudAnimator.SetTrigger("SlideIn");
+
         StartCoroutine(WaitForAnimation());
     }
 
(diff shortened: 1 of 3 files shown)


Assets/Scripts/UI/ObjectiveHudBehaviour.cs
```diff
@@ -72,6 +90,12 @@ public void SetPauseMenuObjective(string objectivePauseTextString)
     /// <returns></returns>
     private IEnumerator WaitForAnimation()
     {
+        yield return new WaitForSeconds(0.5f);
+        foreach (Image h in _objectiveSprites)
```
RyanASwanson, line 94: Could h have a more clear name?

Assets/Scripts/UI/ObjectiveHudBehaviour.cs
```diff
@@ -41,11 +43,27 @@ private void Awake()
     /// <param name="objectiveHudTextString">the text that goes on the ui</param>
     public void ActivateObjectiveHud(string objectiveHudTextString) 
     {
+        //reset to default state if the animation is already going
+        if (_objectiveHudAnimator.GetCurrentAnimatorStateInfo(0).IsName("ObjectiveSlideIn"))
+        {
+            _objectiveHudAnimator.SetTrigger("AbortSlide");
+            _objectiveHudAnimator.ResetTrigger("SlideOut");
```
RyanASwanson, line 50: Not required but a general note is that for animation triggers/bools/etc we like to save them as a const string at the top. This is applicable to this line, the line above, and the one in the if statement

Assets/Scripts/UI/ObjectiveHudBehaviour.cs
```diff
@@ -41,11 +43,27 @@ private void Awake()
...
+            StopCoroutine(WaitForAnimation());
+        }
+
+        foreach(Image h in _objectiveSprites)
```
RyanASwanson, line 54: Could h have a more clear name?

Assets/Scripts/UI/ObjectiveHudBehaviour.cs
```diff
@@ -41,11 +43,27 @@ private void Awake()
     /// <param name="objectiveHudTextString">the text that goes on the ui</param>
     public void ActivateObjectiveHud(string objectiveHudTextString) 
     {
+        //reset to default state if the animation is already going
+        if (_objectiveHudAnimator.GetCurrentAnimatorStateInfo(0).IsName("ObjectiveSlideIn"))
+        {
+            _objectiveHudAnimator.SetTrigger("AbortSlide");
+            _objectiveHudAnimator.ResetTrigger("SlideOut");
+            StopCoroutine(WaitForAnimation());
```
RyanASwanson, line 51: When it comes to StopCoroutine I **highly** recommend saving the coroutine as a coroutine variable when you start it and then when stopping it put in that coroutine variable instead. As multiple of the same coroutine can run at once unity gets confused and just stopping the function doesn't really work.

Assets/Scripts/UI/ObjectiveHudBehaviour.cs
```diff
@@ -72,6 +90,12 @@ public void SetPauseMenuObjective(string objectivePauseTextString)
     /// <returns></returns>
     private IEnumerator WaitForAnimation()
     {
+        yield return new WaitForSeconds(0.5f);
```
RyanASwanson, line 93: For WaitForSeconds we have generally gotten in the habit as a team to save the WaitForSeconds as a variable and set it's value in start/awake. Also could the exact time be a float that can be adjusted?

RyanASwanson (changes requested):
There is some clean up that could be done, overall functionality seems good to me, but I want UX's opinion on having the objective hard-cut to being disabled. Once the clean up is fixed and UX is fine with how it appears it should be good to approve.

Assets/Scripts/UI/ObjectiveHudBehaviour.cs
JeremiahPeters311: the idea is just to have a really short wait because the animator itself is slow and I need this code to run after the animator does its things. making it lower or higher would break the whole thing.

Assets/Scripts/UI/ObjectiveHudBehaviour.cs
JeremiahPeters311: this shouldn't be an issue because it should never be running multiple at once but i went ahead and did this anyway hopefully it doesn't break anything

RyanASwanson (approved):
Still think that declaring the WaitForSeconds at the top is good practice, but aside from that this is cleaned up and works properly. Good work Jeremiah!

PR #926, opened 2025-04-28
gold-eng-Slither-Horror-Moment-Anim-Soft-Locks-Game-LV-2127

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-2127

This task took so long, just because I had to spend time replicating the issue, but I think I finally got it. In this PR there are 2 line changes. In one of them I commented the line out and adjusted a number. In the other I just moved the functionality in the aforementioned line over to a new location. If you uncomment the line you will be able to replicate the issue consistently, I would recommend starting with this. Then afterwards you can test the fixed version. From what I can tell the issue occured because of an order of events thing in which the tentacle mover got destroyed before the tentacle (I don't know how as it is set to happen .1 seconds afterwards but oh well. I guess what would happen is if you skipped over .1 frames you could have it so the tentacle mover destroyed first if it executes first?)

Code Review Estimation: <1 minute


astapay (approved):
Yippee!

PR #929, opened 2025-04-29
GoldQA-Des-Shelfdeath

BreadReaver:
pushed shelf back to prevent players getting caught


pmetcalf2 (commented):
i didn't get stuck at any point around the shelves, go ahead and merge

PR #935, opened 2025-04-29
Gold des first maze hatch adjustments lv 2080

C-Low01:
Swapped hatch location, changed lighting slightly, and added ladder

Note: DO NOT look in scene view ty <3


pmetcalf2 (approved):
hatch makes more sense in the maze, ladder looks good.

PR #937, opened 2025-04-29
Gold des make the chase sequence go through all the hallways lv 2125

BreadReaver:
Adjusted the chase spline so now its one singular chase thing going through every hallway so you are actually being chased the whole time


JamesBondMeow (approved):
looks good, maybe could adjust with chase speed to make it a bit more tense, but overall everything works

MagicTurtle013 (approved):
this is sick

Dusk553 (approved):

PR #943, opened 2025-04-29
gold-eng-4-harpoons-from-refill-LV-2020

Lazydev114:
JiraLink: https://bradleycapstone.atlassian.net/browse/LV-2020
Description: The Ammo rack is counted as a harpoon. It doesn't anymore


RyanASwanson (approved):
Yup this looks good to me. I was only ever able to pick up 3 harpoons and the issue about picking up the harpoon rack is gone. Well done

PR #946, opened 2025-04-29
gold-eng-Gameplay-Settings-Dont-Persist-From-Menu-LV-2128

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-2128

Now I could've simply just adjusted the system in place but I knew that the gameplay settings were saved differently that every other thing in the game. I wanted to fix this. You will find that by playing the game from the main menu, changing the gameplay settings, then pressing new game, the settings will work. Also don't ask me why when setting the Y axis invert in the CameraSettings script I have to do the opposite of the save data. That's how it was before. I guess cinemachine is weird or something?

In Engine Review Estimation: <5 minutes


astapay (approved):
Yippee!

PR #947, opened 2025-04-29
gold-ui-Main-Menu-Animations-Iteration-LV-2061

hhlaxbro:
Jira: https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?assignee=61f6e440c6bd1a00691251d4&jql=assignee%20%3D%2061f6e440c6bd1a00691251d4&selectedIssue=LV-2061

Small changed to the skull animation. Made its landing when touching ground feel more subtle and moving more naturally upon impact.


noodlecat519 (approved):
looks good! i'd sort of like to see some loose dirt swirling up, but that would probably be a post-grad thing if we go through with that, so i'm not withholding approval based on that :)

PR #948, opened 2025-04-29
gold-eng-second-maze-lag-lv-2037

MilesRogers3D:
image

I was able to pretty consistently get rid of any lag spikes on my laptop by tweaking the camera frustum far plane and hiding any weird popping artifacts with distance fog. I also had to make a quick fix on the cel shader that the doors are still using so it could work with unity fog and remove the fog from the monster since it looked weird.

There's a bit of code in here, ExcludeFromCameraFog.cs, which got attached to the cameras in the window renderers to make sure they aren't rendering fog in addition to the fog in the main camera.

The lag spikes happened the worst for me at the beginning of the second maze, so if you want to test a level that would be the best one.


Assets/Scripts/Lighting/ExcludeCameraFromFog.cs
```diff
@@ -0,0 +1,64 @@
...
+/// <summary>
+/// Excludes a camera from rendering fog
+/// </summary>
+[RequireComponent(typeof(Camera))]
+public class ExcludeCameraFromFog : MonoBehaviour
+{
+    /// <summary>
+    /// Initial fog setting in the scene the camera exists in
+    /// </summary>
+    private bool _initialFogSetting = true;
```
RyanASwanson, line 22: Should follow the naming conventions of booleans so should start is words such as is, can, has, does, etc.

RyanASwanson (commented):
I left notes in discord, but I only had 1 comment regarding code

TurquoiseDragon (approved):
No more lag and the fog is a great touch

RyanASwanson (approved):
Looks good to me, left some notes in discord

MagicTurtle013 (approved):

PR #949, opened 2025-04-29
gold-eng-Heartbeat-SFX-bug-LV-2021

Lazydev114:
JiraLink: https://bradleycapstone.atlassian.net/browse/LV-2021
Description: Heartbeat settings were not set, I changed the values, and now it works. The design team might still need to set it to that just right value


RyanASwanson (changes requested):
I found an issue in which the heartbeat sfx would not play after reaching low heath for a second time. For example you reach low health, heal, then reach low health again

RyanASwanson (approved):
Yup! Issue was resolved!

PR #950, opened 2025-04-29
Gold des re add smoke into the second maze generator lv 2081

C-Low01:
Its in there. took me like 4 minutes


SymphonicSweet (approved):
It's there, good job.

PR #951, opened 2025-04-29
gold-art-reduce-datamosh-intensity-LV-2039

MilesRogers3D:
Added a parameter in the material to reduce the intensity of the datamoshing effect and set it to half of what it was before


MagicTurtle013 (approved):

BreadReaver (approved):
Hopefully less motion sickness

PR #952, opened 2025-04-30
gold-eng-Chase-Doesnt-Increase-Player-Speed-LV-2116

astapay:
Done! So there was a little weird thing that was going on in the chase sequence. Basically, the start of the chase itself was setting the player's speed to 0 and resetting it to its original value through just editing the player speed value from the instance of the player movement controller. This caused a little funny business when the story manager tried to change the speed at the same time. Since I basically put the functionality to stop the player movement with the cinematic events, so I just had to move a few things around and cover more things with the cinematic events.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2116?atlOrigin=eyJpIjoiMzBiZjc1ZWRhYTRkNGVkOTg0MDgyMTg4NjY3YmVkNjEiLCJwIjoiaiJ9

Code Review Time: <3 minutes
In-Engine Review Time: <5 minutes (get to the chase scene and make sure the movement speed is correct and the player doesn't move until the cutscene is back. there are some changes to the transform while its happening, but the important part is that you are in the same position as when you enter the cutscene)


Assets/Art/3D/Boss/BossScripts/Attacks/ProceduralAttack/ChaseVineGroup.cs
```diff
@@ -86,16 +84,12 @@ public IEnumerator ActivateThisGroupOfVines()
         {
             //get the player virtual camera
             _playerCam = PlayerCameraController.Instance.PlayerVirtualCamera;
-            _playerMovementController = PlayerMovementController.Instance;
-            _basePlayerSpeed = _playerMovementController.PlayerMovementSpeed;
```
RyanASwanson, line 90: This clean up brings me joy. No changes, just happy 😄

Assets/Scripts/PlayerScripts/PlayerMovementController.cs
```diff
@@ -158,6 +160,32 @@ public void SubscribeInput()
             return;
         }
 
+        if (Camera.main.transform.localPosition != _cameraPositionWithLastInput)
```
RyanASwanson, line 163: I know a similar thing was done in the harpoon gun script was it not? Perhaps we could just have a single event for this kind of thing and put that functionality in the player camera controller? Does require some refactoring though

RyanASwanson (approved):
Functionality looks good and works how it should. Just left 2 comments about how something feels weird about how it is subscribed.

RyanASwanson (approved):
Looks good to me. Learned something out of it too.

PR #956, opened 2025-04-30
gold-ui-quit-confirm-scrim-missing-lv-2050

SolarFoxG:
Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2050

Expected Test Time: 3 Minutes

The scrim wasn't missing, just not prominent. I made it fully block out the background when it's up.


nrice25 (approved):
Approved, there is now a properly screen blocking scrim in the quit pop up

PR #958, opened 2025-04-30
Gold eng fix selected menu buttons lv 2072

nrice25:
[Jira Link](https://bradleycapstone.atlassian.net/browse/LV-2072?atlOrigin=eyJpIjoiZmFmNGViN2IwZTRlNGUwMzg0YTliNTkwNWQ3NTBjNDAiLCJwIjoiaiJ9)
When selecting UI in menus that take you to different pages, it should remember the last one you selected; 
- This is applicable to the main and pause menus

There is currently a bug where the pause menu will fully exit out from the options menu; however, this is no longer a problem in the main menu, and I am unsure of how to fix it; so any help would be appreciated
But, more bugs were removed then created, so ~ it's a net positive :)


hhlaxbro (changes requested):
Nice stuff so far. I think it still needs a bit more tweaking though. I think this will be great for those playing on controller using controller functionality but it's strange for keyboard and mouse users. On keyboard and mouse specifically, menu options should not be visually pre-selected unless their mouse is hovering over an option. The way this works currently is that my mouse can be over Controls, and now both Settings and Controls are highlighted. 

They way it is now will be great for controller functionality, but would it be possible that on keyboard and mouse, there is not pre-selected option?

hhlaxbro (approved):

Assets/Scripts/MenuFunctionality/BackArrowEscapeBehaviour.cs
```diff
@@ -14,62 +14,17 @@
 /// <summary>
 /// allows pressing escape to push a button that this script is attached to
 /// </summary>
-public class BackArrowEscapeBehaviour : MonoBehaviour, IUiSwap
+public class BackArrowEscapeBehaviour : MonoBehaviour
 {
     private PlayerInputMap _playerInputControls;
 
     private Button _backArrow;
```
RyanASwanson, line 21: This variable is no longer in use so can be removed

Assets/Scripts/MenuFunctionality/BackArrowEscapeBehaviour.cs
```diff
@@ -14,62 +14,17 @@
 /// <summary>
 /// allows pressing escape to push a button that this script is attached to
 /// </summary>
-public class BackArrowEscapeBehaviour : MonoBehaviour, IUiSwap
+public class BackArrowEscapeBehaviour : MonoBehaviour
 {
     private PlayerInputMap _playerInputControls;
```
RyanASwanson, line 19: This variable is also no longer in use

Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/UiManager.cs
```diff
@@ -41,6 +117,14 @@ public void SwapInput()
         _onSwapInput?.Invoke();
     }
 
+    private void OnDisable()
```
RyanASwanson, line 120: Even though it is a unity function built in, it would be nice to have an XML

RyanASwanson (approved):
Overall code looks fine, just with some clean up requested. I do like the clean up that was already done about moving stuff out of the BackArrow script. You mentioned a bug in the PR description but I was unable to replicate it, perhaps I need more instructions? I did encounter a minor oversight where if you pressed no on the button for if you want to quit the game, the currently hovered button is continue. Perform the clean up and fix that oversight and this is good by me.

PR #959, opened 2025-04-30
Gold aud master volume doesn't affect chase scene lv 2130

Scottish02:


TurquoiseDragon (approved):
Fantastic audio and everything works

PR #961, opened 2025-04-30
gold-eng-Harpoon-sway-bug-LV-2132

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-2132

Fixed the weapon swap duplicating. The source of the issue seems to have come from a mix up with the _walkingSwayCoroutine and _stopSwayCoroutine variables. The _walkingSwayCoroutine was used to start the stop swap process. Also the coroutines were never set to null in some cases.
How to test: I was best able to replicate the issue before by spamming pause and ads at the same time until the swap functions kept stacking so it happened really fast, though trying other things is appreciated.


TurquoiseDragon (approved):
No more bugs!

astapay (approved):
Yippee!

PR #965, opened 2025-04-30
gold-des-Monster-speed-is-too-fast-LV-2136

JamesBondMeow:
- changed attack interval from 2.5 to 4, which is still faster than before (5.5) but in the middle of the two values


TurquoiseDragon (approved):
Monster feels a lot better now

PR #966, opened 2025-04-30
Gold des UI font doesn't like 4 and 5 lv 2089

pmetcalf2:
Changed the dates on the notes (they all got moved up by 2 days) and changed the demo context text to "1936, three years after the end of prohibition."


SymphonicSweet (approved):
looks good to me, approved

PR #968, opened 2025-04-30
gold-art-window-parallaxing-bug-lv-2119

MilesRogers3D:
Made 2 changes to make sure the fake windows always render correctly:

- Added a global registry for windows, so the closest one to the player can be determined at any time (see: FakeWindowObject.cs, FakeWindowManager.cs)
- Added trees/fog planes to every possible view angle in the capture scene, so if it's pointing the wrong way you don't just see a black box anymore.

This has fixed all bugs with the windows as far as I can tell. It's a little weird with the one on the ceiling but not too weird, the system seems to be able to handle it


Assets/Scripts/Maze/FakeWindowManager.cs
```diff
@@ -103,23 +117,56 @@ public void SetCurrentFakeWindow(FakeWindowObject fakeWindow)
...
-        // Do nothing if fake window is not set
-        if (_fakeWindow.IsUnityNull())
+        // Do nothing if no fake windows are registered
+        if (_fakeWindows.Count == 0)
         {
             return;
         }
         
         // Get global camera position
         var cameraPos = transform.position;
```
RyanASwanson, line 127: Wouldn't it be more efficient to declare variables like this outside of Update that way they don't have to be created and destroyed every frame? This is applicable to all variables declared here.

RyanASwanson (approved):
Approved. Despite that I feel that we really don't need to be declaring variables in update. Feels inefficient to me. Also note the dual window bug I encountered in discord

PR #974, opened 2025-04-30
gold-des-Heartbeat-SFX-Pass-LV-2139

ogNINJAGOLDBOX:
Updated Heart Beat Audio so that the frequency matches the veins pulsing peripheral

https://bradleycapstone.atlassian.net/browse/LV-2139

Just make sure it lines up well. (The audio is a bit quiet, but we don't really have time to fix that I don't believe)


Dusk553 (approved):

TurquoiseDragon (approved):
Sounds good and no obvious bugs

PR #980, opened 2025-05-01
Gold des update subtitle display time lv 2090

pmetcalf2:
updated all the lines i thought needs adjustments to their time to display and their delay before disappearing. Best way i found to test specific lines is to pull a line into the very first beat of the story manager and let it play in the maze scene.


TurquoiseDragon (approved):
Everything looks good

PR #982, opened 2025-05-01
gold-qa-eng-unbreak-the-infinite-ammo-barrel

RyanASwanson:
Reenabled the interactable icon on the infinite ammo barrel. I am aware that the player cannot interact with the infinite ammo barrel in the first maze, that's the whole reason the interactable icons were disabled in the first place. Production decided that isn't getting fixed, they just wanted the issue with reloading fixed.


TurquoiseDragon (approved):
Bug fixed!

PR #983, opened 2025-06-02
pg-eng-Readd-harpoon-wall-sticking-LV-2168

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?jql=assignee%20%3D%206318bb796856bdd60aa045de&selectedIssue=LV-2168

Allows the harpoon to stick into objects that it collides with (This does not apply to enemies)
Basically just turn on infinite ammo and start blasting

In Engine Review Estimation: <7 minutes


nrice25 (approved):
Code looks good and it works in engine; the harpoon does seem to go through a couple of different things although those aren't things we have to worry about with this PR

PR #984, opened 2025-06-02
pg-des-Rearrange-the-windows-in-maze-1/3-LV-2177

JamesBondMeow:
"**There are windows facing in pretty much all directions in the normal maze. That is fine in the other ones, but normally a boat only has windows on it's sides. (Also 2 windows should probably not be visible at the same time, it becomes obviously when the code switches which window we are using to determine the angle)**"

- Adjusted windows in normal maze prefab and copied changes over to BetaMaze and BetaChaseBuildUp prefabs


TurquoiseDragon (approved):
Looks goods, all windows make sense in worldspace

PR #985, opened 2025-06-03
Pg art more people models lv 2148

noodlecat519:
Added four more human models into Assets>Art>3D, where the original human model already is.


pmetcalf2 (approved):
Models look good to me, go ahead and merge

PR #987, opened 2025-06-05
import animation frames

noodlecat519:
Animation frames are in Assets>Art>Animations>Autosave Icon. I have not created an animation clip, because Unity requires you to have a gameobject selected for that and I don't know what object that would be


SolarFoxG (approved):
Sprites for animation frames look good! UI/UX team will take care for implementing the frames properly later.

PR #988, opened 2025-06-06
Pg des make beat 2 play lv 2180

pmetcalf2:
So the spatialized audio in beat 3 (I wrote the task wrong when I made it on the tasksheet) would break the beat and cause none of the dialogue to play, so I just took out the event in the story manager so that everything else plays. TEST THIS IN BUILD, NOT IN EDITOR


MagicTurtle013 (approved):
Shame the spatialized audio had to be removed, but everything else is there so that is better than before lol.

PR #989, opened 2025-06-07
gold-eng-Video-settings-key-control-LV-2070

nrice25:
[Jira Ticket](https://bradleycapstone.atlassian.net/browse/LV-2166?atlOrigin=eyJpIjoiYzIyN2U4MjQwYmUyNGY0ZTg5YWIwYWUzNzcyZTBmZWEiLCJwIjoiaiJ9)

This fixes the navigation for the video section in settings.

Estimated review time: 2 minutes


RyanASwanson (approved):
This looks good to me, works both on controller and keyboard and mouse. Works both when opened from the main menu and from the pause menu. Only thing I noticed which is outside the scope of this task is that you cannot change your resolution without using the mouse.

RyanASwanson (changes requested):
I initially approved this, but I compared it against development and I am a little confused. I am able to navigate the video settings for the most part in development so I'm not certain what has changed. Initially I said the issue with the resolution settings could happen in a separate task but I think I am mistaken, it probably should happen here. If you could provide more information on replicating the initial issue as well as fixing the resolution settings this should be good.

RyanASwanson (approved):
Since you argued the change to the resolution should happen outside of this I will approve of this

PR #990, opened 2025-06-07
pg-eng-Resolution-choices-aren't-consistent-LV-2165

nrice25:
[Jira Ticket](https://bradleycapstone.atlassian.net/browse/LV-2165?atlOrigin=eyJpIjoiYTI2NzhlY2YyMTFhNGZjMWI0NTNjMDEwNGRjN2ZmZDMiLCJwIjoiaiJ9)

Every time the player goes into the settings and video section, the resolution setting should have the correct resolution selected. Need to test in build.

Time Estimate: 10 minutes


RyanASwanson (approved):
Looks good, did leave a note in Discord but I will reiterate here:
Only thing that feels slightly off is that when you do click on the resolution you are always shown the resolutions at the very top. It would feel slightly more natural if you were scrolled down to the resolution you currently have selected.

PR #993, opened 2025-06-10
Pg eng fix the harpoon box lv 2164

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-2164

turns out the reason the harpoon box wasn't working was because of a dialogue trigger hitbox blocking the player interactable raycast. cleo told me it was okay to disable it since he's going to be working on those later anyway so that's what I did. I also increased the size of the harpoon box hitbox because it was pretty small before.


RyanASwanson (approved):
Yup this works in both maze 1 and 3. Granted it already worked in maze 3 but I had to double check. Approved

PR #995, opened 2025-06-10
Pg des implement tree crackle lv 2181

pmetcalf2:
So the tree crackle was already implemented, but someone turned it off, so I just set the folder back to active


Primalwaif5691 (approved):
the sound works

PR #996, opened 2025-06-10
Pg des harpoon rack environmental storytelling lv 2174

Dusk553:
I have moved the harpoons from some racks to be lodged in the wall at points in the game, really only in the area after the swamp room and before the big bins that come up. If you could check that nothing goes wrong on your end for whatever reason, that would be great.

Jira task: https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?jql=assignee%20%3D%2063c32962c52a63dcda83f747&selectedIssue=LV-2174


pmetcalf2 (approved):
Harpoons removed from walls give a bit more sense that the area is more hostile as the player gets further

PR #999, opened 2025-06-11
Monster scream is out of sync lv 2163

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-2163

added creature cry sfx to sfx events
added coroutine to time the sfx playing at the correct moment
fixed broken if statement for playing the start of chase scene sfx (not the cry)

note that currently there will be two cry sounds playing at the start of the chase scene since the old sfx is still baked into the dialogue


Assets/Art/3D/Boss/BossScripts/Attacks/ProceduralAttack/ChaseVineGroup.cs
```diff
@@ -151,12 +158,23 @@ private void CreateMovementAudio()
...
     /// <summary>
     /// Starts playing the movement audio
     /// </summary>
     private void StartMovementAudio()
     {
-        if(!_doesPlayStartAudioEveryChase && !_isFirstChase)
+        if (!_doesPlayStartAudioEveryChase && _isFirstChase)
         {
             //Play the start audio
             RuntimeSfxManager.APlayOneShotSfx?
```
RyanASwanson: So I tested it and I am hearing the screech 3 times at the start of the chase sequence. One time from the dialogue, which obviously we cant do anything about. Once from this line, and once from the line above that you just added in the PlayScreamSfx. We can probably delete the one in here, and have the one in PlayScreamSfx just play the ChaseSequenceStart audio instead, as we shouldn't need a new EventReference for this I believe (Correct me if I'm wrong).
Note **don't** delete the looping audio later in this function, probably just this one.

Assets/General/Audio/FMODScripts/FmodSfxEvents.cs
```diff
@@ -43,6 +43,7 @@ public class FmodSfxEvents : MonoBehaviour
     [field: SerializeField] public EventReference SceneStart { get; private set; }
     [field: SerializeField] public EventReference LimbSpawn { get; private set; }
     [field: SerializeField] public EventReference LimbMove { get; private set; }
+    [field: SerializeField] public EventReference CreatureCry { get; private set; }
```
RyanASwanson, line 46: Based on what I said in my previous comment this can likely go. I believe ChaseSequenceStart is what plays the screech.

RyanASwanson (changes requested):
Currently 3 screeches are playing when there should only be 2.

RyanASwanson (approved):
Just delete the commented out code and feel free to merge!

PR #1002, opened 2025-06-11
PG-UI-hide-level-loading-lv-2153

SolarFoxG:
Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2153

Documentation: https://docs.google.com/document/d/1qyV8T-uRBneLjsCzB8W84C3wD3FJ1SjfEBJIeWANhbo/edit?tab=t.whi3fu8792fm

Expected Testing Time: 5 Minutes

How to Test: Play the game in a BUILD. Pay close attention to the game world after the intro cutscene ends; the game world should remain invisible until objects are loaded in (or for a few seconds during which they should all load in). Afterward, play the game as normal until you reach the first save point, and reload the game using the main menu's Continue button, again checking whether you can observe objects baking into the world or not. For good measure, try to pause the game during the load-in.


hhlaxbro (approved):
As far as I can tell, everything works! Approved

Assets/Scripts/MenuFunctionality/PauseMenu.cs
```diff
@@ -29,16 +30,43 @@ public class PauseMenu : MonoBehaviour
...
+
+    /// <summary>
+    /// Freezes player inputs to give assets a chance to load upon level load
+    /// </summary>
+    /// <returns>Time allowed for loading</returns>
+    private IEnumerator LoadingBuffer()
+    {
+        // Pauses audio
+        FMODUnity.RuntimeManager.StudioSystem.getBus("bus:/In-Game", out FMOD.Studio.Bus masterBus);
+        masterBus.setPaused(true);
```
RyanASwanson: These 2 lines are already in this script. Could we just turn it into a function that takes a boolean? No need to repeat ourself.

RyanASwanson (changes requested):
A bit confused as to why this functionality is specifically in the PauseMenu, but I suppose if it's really needed there it can stay. There is a bug with being able to pause the game and start the dialogue earlier than intended

Assets/Scripts/Managers/GameplayManagers/MainGameplayManagers/GameStateManager.cs
```diff
@@ -104,5 +137,6 @@ public enum EGameplayState
 {
     AboveDeck,
     BelowDeck,
+    Loading,
```
RyanASwanson: Weird to have both an enum and boolean for if the game is loading. Also this enum is so old it is from the first semester, so it can probably "try dying"

Assets/Scripts/MenuFunctionality/PauseMenu.cs
```diff
@@ -141,11 +124,23 @@ private void PauseUIVisibility(bool isVisible,bool shouldToggleAudio)
...
         }
     }
 
+    /// <summary>
+    /// Pauses or resumes all audio. Used whenever the game is paused or given
+    /// a buffer time to load in assets.
+    /// </summary>
+    /// <param name="muteState">Whether the audio should be paused or not.</param>
+    /// <returns></returns>
+    public bool PauseAudio(bool pauseState)
```
RyanASwanson: Why does this return the same boolean it took in completely unchanged?

Assets/Scripts/MenuFunctionality/PauseMenu.cs
```diff
@@ -154,6 +149,17 @@ public void QuitGame()
         Application.Quit();
     }
 
+    /// <summary>
+    /// Toggles a pure white mask to obscure the entire screen.
+    /// </summary>
+    /// <param name="toggleState"></param>
+    /// <returns></returns>
+    public bool ToggleLightScrim(bool toggleState)
```
RyanASwanson: Why is this also returning the exact same boolean it is taking in?

RyanASwanson (approved):
Looks good, just left some tiny notes

PR #1003, opened 2025-06-11
Pg des implement new statue assets lv 2175

MagicTurtle013:
Added the new human models that Caelie found! A majority of the statues just replaced the original guy but I added one new one in the bottle room, you will probably see it.

Just play though normally and notice if any of the models are in the wrong place or note down some cool places where you think I should add one.

Jira Link: https://bradleycapstone.atlassian.net/browse/LV-2175


TurquoiseDragon (approved):
New statues look great and the placements are fantastic!

PR #1004, opened 2025-06-12
pg-eng-Fix-Monster-Clipping-into-walls-LV-2162

astapay:
The monster shouldn't clip in the walls anymore. I wanted to make the attack rotation variable serialized, but I made it a value too big at first and now Unity doesn't like me.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2162?atlOrigin=eyJpIjoiMGRhNzFjN2YwMGI2NDNiOWI0N2ZhYWQ2MTZiMjJlMmIiLCJwIjoiaiJ9

In-Engine Review Time: <5 minutes (check all the monsters and make sure it's at a good range)
Code Review Time: <2 minutes


Assets/Art/3D/Boss/BossScripts/Attacks/ProceduralAttack/ProceduralVine.cs
```diff
@@ -345,9 +346,28 @@ private IEnumerator WSnapAttack()
...
+            {
+                _followTransform.RotateAround(_followTransform.position, Vector3.up, -targetAngle + _MAX_ATTACK_ANGLE);
+            }
+            else
+            {
+                _followTransform.RotateAround(_followTransform.position, Vector3.up, targetAngle - _MAX_ATTACK_ANGLE);
+            }
+        }
+
+        var strikePos = _followTransform.position + _followTransform.forward *
```
RyanASwanson, line 370: Could we maybe get an extra comment or 2?

Assets/Art/3D/Boss/BossScripts/Attacks/ProceduralAttack/ProceduralVine.cs
```diff
@@ -345,9 +346,28 @@ private IEnumerator WSnapAttack()
...
+        var targetAngle = Vector3.Angle(_followTransform.forward, direction);
 
         _followTransform.forward = direction;
+
+        // Attack angle clamping
+        // Basically, we "correct" the follow transform's forward vector by shifting it the difference of
+        // where it is about to go and where we want it to go.
+        if (targetAngle > _MAX_ATTACK_ANGLE)
+        {
+            if (_followTransform.localEulerAngles.y < 180)
```
RyanASwanson: Since the _MAX_ATTACK_ANGLE is a const float could this be as well, just for the sake of consistency? Not required

RyanASwanson (approved):
Yeah the monsters feel a lot better when they attack. I feel the way they slink back into the wall could maybe use some improvements next sprint, but this is a good step to a better feeling monster. Comparing this version directly against development is night and day. Obviously in the busier hallways the monster does clip into some parts of the props, but that's something I will have to talk to design about. Just could use some comments and maybe some consistency with const variables. I did find 2 bugs when looking at this PR, but they are unrelated and are in development. TLDR: Yippee

PR #1008, opened 2025-06-12
new SFX + tweeks to old

Scottish02:


MagicTurtle013 (approved):
New and updated sounds are accessible in Unity

PR #1011, opened 2025-06-18
Pg eng ps4 controller support lv 2229

nrice25:
**REQUIRED**: Ps4 controller
**Estimated time**: 10 minutes to review

This adds in a lightbar effect when the player goes onto the title screen ~ turning it blue
This adds another lightbar effect to reflect the state of the player's health


Assets/Scripts/PlayerScripts/PlayerHealthUI.cs
```diff
@@ -57,6 +60,16 @@ public class PlayerHealthUi : MonoBehaviour
     [SerializeField] [Range(0, 1)] private float _badlyDamagedSaturation;
     [SerializeField] [Range(0, 1)] private float _onDeathSaturation;
 
+    [Header("Health Colors")] 
+    private Color _undamangedColor = Color.green;
+    private Color _lightlyDamagedColor = Color.yellow;
+    private Color _kindaDamagedColor = new Color(.8f,.4f,0f,1f);
+    private Color _badlyDamagedColor = new Color(.8f,.1f,0f,1f);
```
RyanASwanson: This could all probably be a single gradient variable as well, then just set the color to the percentage on the gradient (100% = 100% on the gradient, and so on). I feel that is easier to customize imo. Not a required change but a recommended one

Assets/Scripts/MenuFunctionality/TitleScreenScrolling.cs
```diff
@@ -140,6 +142,11 @@ private void OnEnable()
     {
         _playerInputControls.Enable(); 
         OnUiSwap();
+        
+        if (!DualShockGamepad.current.IsUnityNull())
+        {
+            DualShockGamepad.current.SetLightBarColor(Color.blue);
```
RyanASwanson: Is it possible this could be something that is exposed so it can be easily adjusted?

Assets/Scripts/PlayerScripts/PlayerHealthUI.cs
```diff
@@ -109,27 +136,28 @@ private void UpdateHealthUI(float healthPercent,float currentHealth)
...
-                GlobalColorFilterManager.Instance.Saturation = _onDeathSaturation;
                 break;
         }
         _damagedUIImages[Mathf.Clamp(_damageStatePointer-1,0,4)].gameObject.SetActive(true);
         _heartAnimator.SetInteger(_ANIM_HEALTH_STAGE,_damageStatePointer);
+        GlobalColorFilterManager.Instance.Saturation = _damageSaturations[_damageStatePointer];
+        
+        if (!DualShockGamepad.current.IsUnityNull())
+        {
+            DualShockGamepad.current.SetLightBarColor(_damageColors[_damageStatePointer]);
```
RyanASwanson: These few lines are repeated earlier, putting it in a function would avoid needing to repeat yourself. Not required, but it would be a bit cleaner

RyanASwanson (approved):
Overall this is good with me, but I wish it was more customizable, currently there is no way for design to change it at all.

Assets/Scripts/PlayerScripts/PlayerHealthUI.cs
nrice25: If I did it more than twice I would agree; but I did still do something to make it cleaner :)

RyanASwanson (approved):
Looks good to me. I approve!

PR #1012, opened 2025-06-18
pg-eng-Stuck-harpoons-can-deal-damage-LV-2221

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?jql=assignee%20%3D%206318bb796856bdd60aa045de&selectedIssue=LV-2221

Fixed the issue of stuck harpoons dealing damage. Also removed some old code that wasn't needed anymore.

How to test: Fire the harpoon into a wall the monster would come out of and watch it ignore it. You could also double check that it still deals damage normally. (It does but make sure I'm not crazy)

Code Review Estimation : <2 minutes


Assets/Scripts/PlayerScripts/Harpoon/HarpoonProjectileMovement.cs
nrice25, line 1: That's a lot of code cleanup, time to see if it works

nrice25: It does, the only thing that might still be considered a bug is the harpoon disappears when the enemy collides with it; check with design if this is ok or not

nrice25 (approved):
approved; no longer disappears on monster collision

PR #1013, opened 2025-06-18
pg-eng-Adjust-weapon-sway-LV-2223

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?jql=assignee%20%3D%206318bb796856bdd60aa045de&selectedIssue=LV-2223

The sway of the weapon should feel better than it does on development. Basically just made the math less linear (Animation curves my beloved). Also added functionality for vertical weapon sway as well as it being purely horizontal looked off.

How to test: Walk in a straight line on this branch versus development. It should feel better here.

Code Review Estimation: <8 minutes


Assets/Scripts/PlayerScripts/PlayerCameraController.cs
```diff
@@ -55,21 +55,35 @@ public class PlayerCameraController : MonoBehaviour
...
     private const float _BASE_MOVEMENT_SWAY_SPEED = 0.00005f;
+    private const float _BASE_VERTICAL_MOVEMENT_SWAY_SPEED = 0.0005f;
     private const float _BASE_MOVEMENT_SWAY_INTENSITY = 0.004f;
     private bool _movementSwayRight = true;
     // Cached variables for movement sway
     private float _xSway;
     private float _ySway;
 
+    private float currentSwayDistance;
+    float _swayDistanceLimit;
```
nrice25: Is there a reason to not have these two variables marked as private?

Assets/Scripts/PlayerScripts/PlayerCameraController.cs
nrice25, line 1: I'm gonna try to add an attachment about animation string comparisons. Hopefully it works, I think it might help with the stuttering we see, but maybe it's more of an update change kinda dealio? from yield return null to yield return waitforfixedupdate?
image
Up to you really

nrice25 (approved):

Assets/Scripts/PlayerScripts/PlayerCameraController.cs
RyanASwanson: No you're right. I've been in working in Unreal lately and so I've got accustomed to only needing to declare the accessibility of a section of variables once.

Assets/Scripts/PlayerScripts/PlayerCameraController.cs
RyanASwanson: Just tested changing it to yield return waitforfixedupdate throughout the script and that didn't appear to fix it. I'll try the animation stuff

RyanASwanson: By the looks of it, the image you provided would help with playing animations more efficiently, and I might be able to use something similar for checking what animation is playing if I save the animation and use IsPlaying from the animation, which might be more efficient, but I don't believe it will affect the stuttering. Regardless, it seems more efficient so I'll see what I can implement.

astapay (approved):
Yippee!

PR #1016, opened 2025-06-20
Trying to make the map feel bigger

MagicTurtle013:
The goal of this task was to try to make the map bigger. I took it as make it *feel* bigger and not literally bigger so that is what I attempted to do.

Just play though the game and pay attention to the one-way wall pop ups as they shouldn't be as close to the triggers now. 

I also changed one of the note spots (the flag hallway after the swamp). The idea was to connect more of the map together without changing the player's play space. 

Lmk what you think and if you have more ideas of locations I can do this with because I am having a hard time finding spots.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2245


pmetcalf2 (approved):
additional areas added in some of the offshoots help make the map feel bigger than the area that player that can actually access

PR #1017, opened 2025-06-22
pg-eng-Stuck-harpoons-float-LV-2225

JeremiahPeters311:
https://bradleycapstone.atlassian.net/browse/LV-2225

Harpoons now become children to whatever object they get lodged into, meaning if that object moves, the harpoon moves with it. Easiest to test with the door at the very start of the game.

As a consequence of this, the harpoons would no longer be considered dontdestroyonload, and they would be lost when unloading the scene with the parent object. Since the amount of harpoons is finite, this meant you would be unable to shoot that harpoon ever again. In order to fix this, I made a function that runs whenever a new scene is loaded that checks each of the harpoons and creates a new one if it is missing.

### DIFF
--- a/Assets/Scripts/PlayerScripts/Harpoon/HarpoonGun.cs
+++ b/Assets/Scripts/PlayerScripts/Harpoon/HarpoonGun.cs
@@ -757,6 +757,7 @@ public class HarpoonGun : MonoBehaviour
         for (int i = 0; i < _harpoonPoolingAmount; i++)
         {
             ObjectPoolingParent.Instance.InitiallyAddObjectToPool(_harpoonSpearPool[i].gameObject);
+            _harpoonSpearPool[i].gameObject.SetActive(false);
         }
     }
 
--- a/Assets/Scripts/PlayerScripts/Harpoon/HarpoonGun.cs
+++ b/Assets/Scripts/PlayerScripts/Harpoon/HarpoonGun.cs
@@ -2,7 +2,7 @@
 // File Name :         HarpoonGun.cs
 // Author :            Tommy Roberts
 // Contributors:       Ryan Swanson, Adam Garwacki, Andrew Stapay, David Henvick, 
-//                     Miles Rogers, Nick Rice, Charlie Polonus
+//                     Miles Rogers, Nick Rice, Charlie Polonus, Jeremiah Peters
 // Creation Date :     9/22/2024
 //
 // Brief Description : Controls the basic shoot harpoon and retract functionality.
@@ -16,6 +16,7 @@ using Cinemachine;
 using Unity.VisualScripting;
 using PrimeTween;
 using FMOD.Studio;
+using UnityEngine.SceneManagement;
 using UnityEngine.Rendering.Universal;
 
 /// <summary>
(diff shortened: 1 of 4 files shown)


Assets/Scripts/PlayerScripts/Harpoon/HarpoonGun.cs
```diff
@@ -200,6 +201,8 @@ private void SubscribeToEvents()
         CameraManager.Instance.GetOnCinematicStartEvent().AddListener(HideReticle);
         CameraManager.Instance.GetOnCinematicStartEvent().AddListener(ResetFocus);
         CameraManager.Instance.GetOnCinematicEndEvent().AddListener(ShowReticle);
+        
+        SceneManager.sceneLoaded += StartRepairHarpoons;
```
RyanASwanson: Damn Unity doesn't follow our naming conventions. Smh. Anyway carry on, this isn't a real comment

Assets/Scripts/PlayerScripts/Harpoon/HarpoonGun.cs
```diff
@@ -740,6 +745,38 @@ private void CreateInitialHarpoonPool()
         }
     }
 
+    /// <summary>
+    /// starts the RepairHarpoons coroutine
+    /// it won't let me start a coroutine directly from the scene loaded event, so i did this work around
```
RyanASwanson: This actually is possible through delegates, so you can subscribe the coroutine directly, problem is I have no idea how to unsubscribe the delegate, so this is fine to ignore.

RyanASwanson (changes requested):
I tried firing a harpoon and returning to the main menu, then pressing new game and I got an error that prevented me from playing. I think you are on the right track with this one. There is an event in the AframaxSceneManager called _onBeforeSceneChange, this is called at the start of changing a scene. You could subscribe the harpoons to child themselves back to being children of the Object Pooling Parent with the AddObjectAsChild function, then to disable themselves. You could also repeat thing in the MazeSubSceneManager that Miles made. Perhaps make a new event in that script and subscribe the harpoons to do what I previously mentioned. I would probably call the event in the DestroyMaze() function but that is just a guess and it might be more fitting somewhere else. If we do that we probably won't even need the functionality to repair harpoons.

Assets/Scripts/PlayerScripts/Harpoon/HarpoonGun.cs
```diff
@@ -746,34 +750,13 @@ private void CreateInitialHarpoonPool()
...
-        //check and fix each harpoon that needs fixed
         for (int i = 0; i < _harpoonPoolingAmount; i++)
         {
-            if (_harpoonSpearPool[i] == null)
-            {
-                _harpoonSpearPool[i] = Instantiate(_harpoonPrefab,
-                    _playerLookDirection.position, Quaternion.identity).GetComponent<HarpoonProjectileMovement>();
-                _harpoonSpearPool[i].gameObject.SetActive(false);
-            }
+            ObjectPoolingParent.Instance.InitiallyAddObjectToPool(_harpoonSpearPool[i].gameObject);
```
RyanASwanson, line 759: Also probably should set the object active to false or at least add a toggle to do so.

RyanASwanson (approved):
Looks good, just left a single comment about disabling harpoons. Aside from that we are good. This actually should fix a bug that was intended for next sprint I believe.

Assets/Scripts/PlayerScripts/Harpoon/HarpoonGun.cs
JeremiahPeters311: i thought about that, but from my testing it looks like the object pooling parent does that already

PR #1018, opened 2025-06-23
pg-eng-mouse-visability-during-controller-gameplay-LV-2218

astapay:
Mouse now disappears during controller gameplay for good! I kinda thought "why don't we just get rid of the mouse right away when we toggle the controller since the keyboard controls still work no matter the setting?" so this is the result of that. The original task was to just keep it gone when using controller during cutscenes and note reading. Hopefully everything still works out.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2218?atlOrigin=eyJpIjoiNGExMTc0ODJiZjg5NGM5OGJjNTBhZDhhM2UxM2RjNjUiLCJwIjoiaiJ9

Code Review Time: <2 minutes
In-Engine Review Time: <10 minutes (play through the game with both keyboard and controller and see if everything is fine in terms of when the mouse appears)

### DIFF
--- a/Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/UiManager.cs
+++ b/Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/UiManager.cs
@@ -113,7 +113,20 @@ public class UiManager : MainUniversalManagerFramework
     {
         _isUsingController = !_isUsingController;
         SaveManager.Instance.GetGameSaveData().IsUsingController = _isUsingController;
+        ToggleMouse();
         _onSwapInput?.Invoke();
+    }
+
+    private void ToggleMouse()
+    {
+        if (_isUsingController)
+        {
+            UnityEngine.Cursor.lockState = CursorLockMode.Locked;
+        }
+        else
+        {
+            UnityEngine.Cursor.lockState = CursorLockMode.None;
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/UiManager.cs
```diff
@@ -113,9 +119,22 @@ public void SwapInput()
     {
         _isUsingController = !_isUsingController;
         SaveManager.Instance.GetGameSaveData().IsUsingController = _isUsingController;
+        ToggleMouse();
         _onSwapInput?.Invoke();
     }
 
+    private void ToggleMouse()
```
RyanASwanson, line 131: Xml pwease

RyanASwanson (approved):
For the most part it seems good. Left a single code comment about an XML. I played this for a while in build and the only time the mouse appeared when I didn't expect it to was when I opened the settings page. Currently the settings page can't be fully accessed by controller as the resolution cannot be changed so this makes sense, but maybe add a toggle so once that issue is resolved we can switch it? Otherwise I found no issues, just fix those and we are good. Though it wouldn't surprise me if there is some super secret combination of random stuff you can do to make the mouse appear that I missed, but I suppose we shall see in the QA time

RyanASwanson (approved):
Yup this looks good, would prefer if it had a toggle, but aside from that this is fine with me

PR #1020, opened 2025-06-24
pg-eng-Steam-Setup-LV-2216

astapay:
So I looked up a tutorial and this is as far as I can go without having access to the Steamworks page (I think). lmk if I have to do anything else.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2216?atlOrigin=eyJpIjoiOTczNzMzMWJkY2U0NDFlZmE3OTE1NWRlNjhjOWEzNjQiLCJwIjoiaiJ9

Code Review Time: <10 minutes (everything you see here was added by an open source unity package. Basically just make sure all the right files are there, if there are any that come to mind that should be here
In-Engine Review Time: <10 minutes (make sure the game still runs)


RyanASwanson (approved):
Looks good, but I would like some clear documentation for unlocking achievements.

Assets/Scripts/Steamworks.NET/SteamAchievements.cs
```diff
@@ -0,0 +1,125 @@
...
+{
+    /// <summary>
+    /// Container class for all information related to an achievement
+    /// VARIABLES PASSED INTO THE CONSTRUCTOR MUST MIRROR THE INFORMATION ON THE STEAMWORKS ACHIEVEMENT
+    /// CONFIGURATION PAGE
+    /// </summary>
+    private class AchievementContainer
+    {
+        public Achievement AchievementId;
+        public bool Achieved;
```
RyanASwanson, line 25: Should start with words such as is, has, does, etc. as it is a bool. Though I am not sure how much of this is stuff that came from a package or that you made, so if it is from a package it's probably fine

RyanASwanson (approved):
Looks good, I appreciate the documentation

PR #1021, opened 2025-06-24
pg-des-Statue-Phasing-LV-2251

pmetcalf2:
added a box collider to the statue that people were reporting that they were able to pass through.


MagicTurtle013 (approved):
scare has been fixed and the man has a collider

PR #1022, opened 2025-06-24
Pg art monster vines lv 2230

noodlecat519:
Vine kit can be found in Art>3D>VineModularKit. There's also an art test gym called VineKitImport if you want to see them in a scene (although they're all stacked, I was just testing pivot points)


pmetcalf2 (approved):
Looks good, vines will help make the areas look more infested

PR #1025, opened 2025-06-25
Pg des the ruth decal incorrect projection lv 2250

pmetcalf2:
Pushed all the Ruth decals into the wall that had their projectors extended out. LMK if any decals pop up on the gun.


MagicTurtle013 (approved):
I was not able to see the decals on my gun

PR #1028, opened 2025-07-01
pg-eng-Music-in-cutscene-drowns-out-dialogue-LV-2271

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?customFilter=ari%3Acloud%3Ajira-software%3A792ff6f6-f5ba-4734-a1cd-457d34518a06%3Acustom-filter%2Factivation%2F489b3b11-a1a5-4c9e-afed-c8d621aa5e68%2F32%2F0ffdcf80-e13c-47e4-8c44-6b732fc8693c&selectedIssue=LV-2271

Okay so I could've sworn I was going crazy while fixing this for a bit. I disabled all music from playing just to test things, and then I still hear music in the cutscenes. As it turns out the music played in cutscenes is played by the Cinematic Manager and not the persistent audio manager. I added a few things to give options in clearing this up. I gave the cinematic manager a toggle to not play any audio, as well as a volume slider. For now I just reduced the volume to 25% for all the cutscenes, but now it is easier to customize audio for cutscenes.

How to test: Basically just press play on the game and for the demo context and intro cutscene the music should be really quiet. Feel free to toy with this if you want and adjust the slider to check it's working properly.


Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/PersistentAudioManager.cs
```diff
@@ -356,6 +356,36 @@ public void StartMusicByID(int id)
         StartMusicByReference(FmodPersistentAudioEvents.Instance.MusicInGame[id]);
     }
 
+    /// <summary>
+    /// Stops any current music track actively playing
+    /// </summary>
+    public void StopCurrentMusicTrack()
+    {
+        // If there is music already trying to fade in, stop doing that
+        if (_musicFadeInCoroutine != null)
```
nrice25: Can multiple of these happen at the same time? Because it could probably be an if else-if chain so then it wouldn't need to check each one
Also, ! .IsUnityNull()

nrice25 (approved):
Works in engine. Code is good with some cleanup that can happen. I thought that the audio slider was for the background sfx and music, not just the music. So I tested a lot more than needed.

Assets/Scripts/Managers/UniversalManagers/MainUniversalManagers/PersistentAudioManager.cs
RyanASwanson: No the _musicFadeInCoroutine should never have multiple running at the same time. Though I'll double check in engine to see if it's ever possible. But yeah you are right about the .IsUnityNull

RyanASwanson: I realized I misinterpretted what you said. I thought you were referring to multiple of this specific coroutine running at once, but you meant multiple of the 3 if statements. The first 2 can be an if else, the last one should probably remain an if that is seperated

PR #1029, opened 2025-07-01
pg-eng-Harpoons-will-not-appear-on-the-ground-or-ceiling-LV-2269

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?jql=assignee%20%3D%206318bb796856bdd60aa045de&selectedIssue=LV-2269

Though this likely also probably fixes the following as well.
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?jql=assignee%20%3D%206318bb796856bdd60aa045de&selectedIssue=LV-2273

Harpoons now properly stick into the ceiling and floor. Reason why that happened was the harpoon got confused where to stick itself because the colliders for ceilings and floors of each room were on the same object, and the harpoon would stick into the first collider that it found on that object. This was fixed in the line for calculating the closest collision point in HarpoonProjectileMovement to just use the 'other' variable. NOTE: Some ceilings don't have colliders so they don't stick into every ceiling. I'm planning on talking to design about that part.
Also fixed an issue with harpoons not sticking into objects that have an unusual scale and the harpoon inheriting that scale. Now it only childs itself to the door, which is the only moving object the harpoon can stick to.

How to test: Turn on infinite ammo with the dev console and just start blasting.


Assets/Scripts/PlayerScripts/Harpoon/HarpoonProjectileMovement.cs
```diff
@@ -113,8 +119,10 @@ public void ImpaleHarpoon(Collider other)
         }
         
         _harpoonDamage.CanApplyDamage = false;
+        
+        // Finds the closest point on the collider to stick to
         transform.position = Physics.ClosestPoint(transform.position,
```
nrice25, line 124: Huh, closest point sounds a lot more elegant than the convoluted system I made for vfx. Half of me is wondering if we could make the harpoon travel a certain distance of it's own length; granted, we would then need to know the length of the harpoon, and I don't think that's easily accessible? At least off the top of my head. I could look that up later. Granted, this is probably another task in and of itself. Regardless, good code.

nrice25: We could then tie how far it goes based on the material it collides with...
But that's too much. Granted, the current system does take into account the current polygon the harpoon collides with. So, it's kind of doing some similar things already?

Assets/Scripts/PlayerScripts/Harpoon/HarpoonProjectileMovement.cs
nrice25, line 1: Overall, code looks good; I will hopefully test it tomorrow; I should be asleep right now.

nrice25 (approved):
Approved, no major notes for code and it works in game.

PR #1030, opened 2025-07-01
fixed obvious lighting issues in BetaMaze

MagicTurtle013:
There was a lighting issue on the ceiling tile before the final right turn in BetaMaze before the spanner.

I added and modified reflection probes throughout the entire level and re-baked them. 

Just play the game, lighting quality should be generally improved in BetaMaze and the issue of the ceiling tile is also gone.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2294


pmetcalf2 (approved):
Lighting looks better, and the roof tile has been fixed.

PR #1031, opened 2025-07-01
Player should no longer be able to cheese chase sequence

MagicTurtle013:
Player used to be able to break the monster in the chase sequence by standing at the back of the first straight hallway.

Moved some props to block corners that the player could stand in to avoid the monster.

Just play the chase sequence and try to stand in corners to avoid the monster. If you find any spots lmk so I can fix them.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2296


pmetcalf2 (approved):
player can no longer cheese the chase monster, and also changed collider helps with potential weirdness of the player sometimes being missed

PR #1032, opened 2025-07-02
Pg art update monster scratches decal lv 2187

noodlecat519:
You can find a decal projector using the new scratches in DecalMazeTesting art test gym, or you should be able to just play the game and look for scratches as it will update all of them


pmetcalf2 (approved):
Looks great in scene, i like the almost wateryness of the scratches

PR #1033, opened 2025-07-02
messed with the tree roots to add visibility to monster animation

MagicTurtle013:
Tree roots were blocking the player's camera when playing the first monster animation in the cutscene.

Just play through the game till you get to the part and lmk what you think.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2295


pmetcalf2 (approved):
Branches can no longer block the players view when the monster animation plays.

PR #1034, opened 2025-07-03
Pg des fix ceiling tile colliders lv 2302

MagicTurtle013:
Added colliders onto ceiling tiles that didn't have them prior.

Original reason for this PR was because harpoons would go through the ceiling and it was thought to be a collider issue. It turned out to actually be the vfx manager didn't have the ceiling and floor materials in the list of materials that give off impact vfx, so adding them to the manager fixed it.

Shoot the ceiling and floor, they should have impact vfx with the harpoon sticking in with the a dent decal.

Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2302


pmetcalf2 (approved):
Ceilings are now able to have harpoons shot into them and there aren't any missing ceilings

PR #1040, opened 2025-07-08
pg-eng-Empty-Ammo-animation-cancel-LV-2304

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?jql=assignee%20%3D%206318bb796856bdd60aa045de&selectedIssue=LV-2304

The issue happened when restocking when you had exactly 1 ammo remaining, the trigger for reloading from empty wrongly went off, then the next time you ran out of ammo that trigger would cause you to skip the empty animation.
The issue should be fixed just by the changed if statement but I threw in a reset trigger just to be certain that it is never possible to happen.

How to test: Fire harpoons and restock at various ammo remaining. Then on running out of ammo the correct animation should play

In Engine Review Estimation: <7 minutes


nrice25 (approved):
Code looks good and it works in engine, approved.

PR #1050, opened 2025-07-14
pg-eng-Player-cannot-move-after-pressing-continue-LV-2256

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?jql=assignee%20%3D%206318bb796856bdd60aa045de&selectedIssue=LV-2256

Player can move now. Problem happened because recently functionality was added to essentially pause the game until the game is done loading. The issue is that the game unsubscribed input when pausing (which is intentional) and never resubscribed it. I made it so that it doesn't attempt to unsubscribe or subscribe until the level is fully loaded. To answer the question of "Why not use the _isGameLoading", which is the one right next to the new bool I made, that doesn't work due to the order that things happen in. The game is paused, then sets that bool to true, and I cannot change that order without destroying things.

How to test: Play the game up to save points and try loading the game from those points.


Assets/Scripts/Managers/GameplayManagers/MainGameplayManagers/GameStateManager.cs
```diff
@@ -83,6 +84,7 @@ private IEnumerator LoadingBuffer()
         _isGameLoading = false;
         PauseMenu.Instance.ToggleLightScrim(false);
         PauseMenu.Instance.PauseToggle();
+        IsGameLoadingCompleted = true;
```
nrice25, line 87: I never see this set to false?

nrice25:
The player sees the void after clicking continue and when the player goes back to the pause menu they no longer have the ability to use their mouse.

Assets/Scripts/Managers/GameplayManagers/MainGameplayManagers/GameStateManager.cs
RyanASwanson: The variable defaults to false when declared, and then sets itself to true and never returns to false until the scene is entirely reloaded

nrice25 (approved):
Could not seem to replicate the only issue I had with this PR. It works and the code looks good, approved.
The only thing is that we don't have a standard for internal variables

PR #1051, opened 2025-07-14
pg-eng-Monster-Slither-Camera-Issue-LV-2305

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32?jql=assignee%20%3D%206318bb796856bdd60aa045de&selectedIssue=LV-2305

Okay so I am going crazy because this works and I don't get it. I worked on this task like 3 hours last sprint and now 1 hour this sprint and somehow I got it to work. How you might be asking? I disabled the Virtual Camera on the cinematic, and replaced it with a practically identical one, that worked for some reason, so I copied the component of the new Virtual Camera and applied it to the old one, and somehow it still works. Witchcraft. I swear this was an issue with cinemachine/unity because I tried adjusting values on the original Virtual Camera and it would only apply after a few times of attempting. So I'm going to blame cinemachine and move on with my life.

How to test:
Play the game up to the cinematic with the monster slithering away. The camera should follow the monster properly now. Make sure I didn't break anything else in the process


nrice25 (approved):
Approved, works in engine and doesn't break anything else from what I could tell.

PR #1052, opened 2025-07-20
pg-eng-gameplay-settings-sfx-bug

JeremiahPeters311:
you can read my original rant/explanation here https://discord.com/channels/1366544497927782501/1367536366384189460/1396409363970916424

but basically, every individual option in the settings menu is set to the current saved value when it is first enabled each time the settings scene is loaded. as there is no way to update the values quietly, this meant the click sound was playing when this happened. this pr mainly exists to update the click sfx script to avoid playing under certain circumstances, and other various changes were made for parity. 

This includes:
moving GameplaySettings.cs to the same object that VideoSettingsVehaviour.cs and AudioSettings.cs are on since they all serve the same purpose

moving the lines setting the values of the toggles inside GameplaySettings.cs to awake instead of onEnable (I thought this would fix things but it didn't and left it)

reworking the invert x and y sliders to use onValueChanged to update saved settings instead of manually creating listeners in code

Also because of the changes made to ButtonSFXManager I had to create another function that bypasses the check I created in order to ensure that the title screen works properly


Assets/Scripts/UI/ButtonSFXManager.cs
```diff
@@ -16,11 +16,28 @@ public class ButtonSFXManager : MonoBehaviour
...
     /// </summary>
     public void PlayClickSFX()
+    {
+        if (_startTime != 0)
+        {
+            RuntimeSfxManager.APlayOneShotSfx?.Invoke(FmodSfxEvents.Instance._buttonUsedSFX[_clickSFXIndex], Vector3.zero);
+        }
+    }
+
+    public void AlwaysPlayClickSFX()
```
nrice25, line 39: To be picky, please add function headers :)

RyanASwanson (approved):
Looks good, just add the XML as Nick mentioned

nrice25 (approved):
XML comments added and it works in engine, approved

PR #1054, opened 2025-07-21
pg-ui-player-can-pause-during-opening-white-fade-in-LV-2284

SolarFoxG:
Jira Issue: https://bradleycapstone.atlassian.net/browse/LV-2284

Documentation: [Pause Menu](https://docs.google.com/document/d/1qyV8T-uRBneLjsCzB8W84C3wD3FJ1SjfEBJIeWANhbo/edit?tab=t.whi3fu8792fm#heading=h.z8cyow293u9f)

Time to Test: 10 Minutes

How to Test: When the player loads into the maze scene, either from a new game or by continuing, the screen will have a white border quickly retract itself before the screen is fully unobscured. If you pause the game as the screen is revealing itself, the transition should disappear entirely. As this has to do with checking game states, it may be worth it to run through the game to test this functionality at each save point by launching the maze from Continue on the main menu.


hhlaxbro (approved):
nice

nrice25 (approved):
Works in engine and code looks good, approved

Assets/Scripts/UI/SceneTransitionBehaviour.cs
```diff
@@ -40,4 +40,28 @@ public void PlayTransition(string animationTrigger)
...
     }
+
+    /// <summary>
+    /// Checks whether a transition is currently in play.
+    /// </summary>
+    public bool CheckIfTransitionIsActive()
+    {
+        if (_transitionImage.enabled)
+        {
+            return true;
```
RyanASwanson: We can simplify this function down to 
return _transitionImage.enabled;
no need for an if else

Assets/Scripts/MenuFunctionality/PauseMenu.cs
```diff
@@ -61,6 +61,21 @@ private void CheckSingletonInstance()
...
     }
 
+    /// <summary>
+    /// Checks if a scene transition is currently animating. If so, the animation is disabled. 
+    /// Regardless, proceeds to check whether the game's pause state can be toggled or not.
+    /// </summary>
+    private void PauseAnimCheck()
+    {
+        // Disables scene transition if it's active
+        if ((SceneTransitionBehaviour.Instance.CheckIfTransitionIsActive() == true) && !GameStateManager.Instance.IsGameLoading())
```
RyanASwanson: We can simplify the first half of this if statement down to 
(SceneTransitionBehaviour.Instance.CheckIfTransitionIsActive())
we don't need to have the '== true' at the end

RyanASwanson (approved):
Works fine in engine, code works fine functionality wise, but some parts could be written shorter.

PR #1055, opened 2025-07-21
Import new versions + fixes

Scottish02:


pmetcalf2 (approved):
went through and played all the new sound effects and they were there, monster music stops playing once you kill a monster, looks good to me.

PR #1064, opened 2025-07-25
pg-eng-Fix-Harpoons-Floating-In-Air-LV-2348

RyanASwanson:
https://bradleycapstone.atlassian.net/jira/software/projects/LV/boards/32/backlog?assignee=6318bb796856bdd60aa045de&selectedIssue=LV-2348

Ended up finding and fixing a separate issue in this one as well
For the main issue on rare occasions the harpoons would stick in the air when hitting enemies. This is really hard to replicate and I didn't have an effective method other than just constantly shooting monsters
For the second issue I discovered that monsters don't actually have a hitbox for a second after spawning. This means 2 things, for one the monster can't deal damage for 1 second, which is a good thing, but it also can't take damage for 1 second. I split it into 2 separate hitboxes, one for dealing damage and one for taking it. It now spawns the hitbox for taking damage instantly and enables the second after a delay.

How to test: Try shooting monsters to see if the harpoon sticks in air on hitting them.
Try shooting monsters within a second of them spawning, they should die properly now
Try taking damage from a monster as soon as it spawns, it should still have a 1 second delay before dealing damage

In Engine Review Estimation: <10 minutes

### DIFF
--- a/Assets/Art/3D/Boss/BossPrefabs/Stem(wahckaMole2).prefab
+++ b/Assets/Art/3D/Boss/BossPrefabs/Stem(wahckaMole2).prefab
@@ -745,7 +745,7 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 5268b2c91a0bee146a8862a8ded302c2, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  _delayFirstSpawn: 1
+  _delayFirstSpawn: 0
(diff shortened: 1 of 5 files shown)


nrice25:
Code looks good, gonna check ~~in engine now~~ eventually;
I forgot the steam fix isn't in every branch.

nrice25 (approved):
Approved, works in engine and code looks good.

PR #1065, opened 2025-07-28
pg-eng-Steam-Achievements-not-appearing-in-game

astapay:
Hooray! I think this is all I needed to do (according to the forum that Andrea sent). Not really sure how I can test this without being able to run this through the Steam build, but we can try I guess :D.

Jira Task: Not up on Jira yet

Code Review Time: <1 minute
In-Engine Review Time: I really have no idea. I tried making a build just through Unity and nothing really happened, so I think this just has to be merged and tested through Steam to be tested properly.


RyanASwanson (approved):
Yeah that looks good. Can't really test if the actual functionality does what it should until the next steam build, but it doesn't break anything in engine as far as I can see.

PR #1069, opened 2025-07-29
No more pop

Scottish02:


MagicTurtle013 (approved):
no more popping sound in the main menu

PR #1070, opened 2025-07-30
pg-ui-cutscene-reimplementation-LV-2214

SolarFoxG:
Jira Task: https://bradleycapstone.atlassian.net/browse/LV-2214

How to Test: In intro and outro cutscenes, screen resolution should allow for cinematics to be shown in full in almost every resolution. Test the intro and outro scenes in various resolutions to ensure that, in particular, subtitles are always visible.

Expected Testing Time: 10 Minutes

Slightly added more coverage to the cutscenes animatics. Did so by altering the general scaling of the cutscene in relation to the screen (giving it generally 25% padding on normal displays to account for some taller displays) and changing the background seaweed to provide more coverage for white space.


MagicTurtle013 (approved):
Went ahead and made a build and ran through almost every resolution, the videos are cleanly scaled to each resolution and I could always read the subtitles.

PR #1073, opened 2025-07-30
flickers should be working properly now

MagicTurtle013:
There were some places in the map that had a light flicker but nothing happened. I removed the flickers that did nothing and slightly improved the slytherin horror moment flicker to look better. 

Just play and you shouldn't hear random flicker sfx without lights changing. Currently the only flicker is the slytherin horror moment, all other flickers are ambient and not activated via a trigger.

Jira: https://bradleycapstone.atlassian.net/browse/LV-2355 atlOrigin=eyJpIjoiMmRhYmJhYmNiODc2NDI4OGFkYmU2Mjk0ZTY4ZjFiM2IiLCJwIjoiaiJ9


pmetcalf2 (approved):
Flicker is only tied to the lights and slytherin now, go ahead and merge when updated from main

PR #1080, opened 2025-08-08
Added Placebo Entertainment Logo to Intro Credits

MagicTurtle013:
Added the Placebo Entertainment Logo to the Intro Credits


RyanASwanson (approved):
Yeah looks good. Weird nitpick but for some reason it feels like the logos don't remain on screen for the same amount of time. Like the Fmod logo isn't there for too long but the Placebo logo is. Though I didn't think that was something you could even change so perhaps it's just a quirk of how Unity does it's splash screens? Regardless that's no reason to not approve.

PR #1081, opened 2025-08-08
pg-eng-fix-credits-scroll

RyanASwanson:
https://docs.google.com/spreadsheets/d/16URqM0YXxcVfVZ6MlO4QJtvad7bngXEr7y3WLCQJBXY/edit?gid=387920552#gid=387920552

The credits should no longer be able to get stuck and not complete. Also while I was at it I fixed the issue of the screen scroll speed being depended on resolution. I would like someone on the UI/UX team to verify this is as it should be and is scaling correctly depending on resolution.

How to test: Play through the credits multiple times to be sure it never gets stuck (Best way to replicate this is to change the size of the game view while the credits were scrolling), to make sure that the scroll speed is consistent across resolutions, and to make sure that the screen scaling didn't get messed up in the process.


hhlaxbro (approved):
nice

PR #1083, opened 2025-08-13
pg-des-trigger-readjustments

ogNINJAGOLDBOX:
https://docs.google.com/spreadsheets/d/16URqM0YXxcVfVZ6MlO4QJtvad7bngXEr7y3WLCQJBXY/edit?gid=387920552#gid=387920552

Rows 3-4

Adjusted the Trigger at the start of the Chase so that the dialogue lines up with the cutscene.

Swapped the placement of the Boat Bell Toll and Line 34 about the boat as it felt it would make more sense for the boat to make the sound before the player comments.

I wanted to originally adjust all the Boat Associated SFX and VFX to be less choppy, for example the Distorted Visual with the Boat Crash being lined up, but upon messing with it, I felt if they were too close the Distortion SFX gets muted out by the boat crash. Not sure if there is a way to do this without extra code and assigning it all to 1 trigger, which I don't think is worth the time. My only concern is if a player goes slowly, they will get the Distortion and Boat overturning before the Boat Crash SFX, so it won't make much sense

Edit: Added another change which was just adjusting the Exit latch collider for the chase end. I almost died trying to hit it like 5 times and decided to quickly make the collider larger

To Test: Just test out the Boat Bridge Triggers and Chase Triggers (Edit: And End Game Latch Collider) to make sure they are lined up well

### DIFF
--- a/Assets/Prefabs/WorldPrefabs/Mazes/Beta/ActualMazes/BetaChaseSequence.prefab
+++ b/Assets/Prefabs/WorldPrefabs/Mazes/Beta/ActualMazes/BetaChaseSequence.prefab
@@ -58758,8 +58758,8 @@ BoxCollider:
   m_ProvidesContacts: 0
   m_Enabled: 1
   serializedVersion: 3
-  m_Size: {x: 0.00868223, y: 0.011627636, z: 0.002734126}
-  m_Center: {x: -0.00000011920929, y: 0.0000045979023, z: 0.0012698083}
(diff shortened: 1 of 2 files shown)


RyanASwanson (approved):
I like the change to the chase trigger, feels more natural. The boat trigger I feel a tad mixed on though, I felt that the bell toll happening later leads to the player turning around and not missing the boat going underwater.

MagicTurtle013 (approved):

PR #1085, opened 2025-09-04
fixed ceiling tile

MagicTurtle013:
fixed the ceiling tile seam in the first monster encounter


RyanASwanson (approved):
Yup looks good. Checked both in editor view and in gameplay and I couldn't manage to find a gap.

PR #1087, opened 2025-09-05
Branching Path Fix

pmetcalf2:
Moved the right door up and its trigger on the last crossroads by the spanner. Just run to the end, take the right path at the last crossroads, and let me know if there are any issues or concerns.


RyanASwanson (approved):
Approved

PR #1088, opened 2025-09-23
Tiny map adjustments

MagicTurtle013:
All of the fixes were so small that I just put them in here.
- Fixed clipping through the hatch at the beginning
- Fixed the player spawn to it is now centered
- Fixed the generator alignment after using the spanner
- Slightly increased the reload sfx delay from 3.41 -> 3.725


RyanASwanson (approved):
Looks good to me. The player is aligned at the start. The player seems to be unable to clip their camera into the hatch (though the weapon does still slightly clip, but that might be unavoidable). I was unable to notice a difference in the generator location before and after using the spanner. The reload is better aligned but I think it could still maybe be a tiny bit more delayed, like by 0.1-.2 seconds? Regardless this is approved. I don't need to rereview to check the reload thing

